PR #5, opened 2018-10-11
Create 03-out-of-band-nfs

ssalaues:
Initial design proposal for NFS/SOFS out of band ingestion.


03-out-of-band-nfs.md
```diff
@@ -0,0 +1,108 @@
+# ZIP#3 (Zenko Improvement Proposal) Out-of-Band NFS Metadata Ingestion
+
+## Overview
```
ssalaues: Make existing data visible to Zenko

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,108 @@
...
+
+This document is a proposal for a new feature that allows for metadata ingestion of filesystem data through NFS mounts. It will continuously sync both pre-existing and newly created files on a filesystem, and allow their management through Zenko.
```
vrancurel: Use listing, attributes and content, hierarchy but not metadata which is too vague.

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,108 @@
...
+## Alternatives
+
+#### Mount the NFS export directly into the pods.
+
+- **Option 1: Requires special privileges**
+  With SYS_ADMIN or similar capabilities applied to each of the cloudserver containers, they would be able to directly “mount” to an NFS directly which would greatly simplify configuration. However, this opens the door to potentially severe security risk because cloudserver is an externally exposed service (in fact, the only one).
+
+- **Option 2: A userspace NFS client**
+  With a userland NFS client, such as [nfusr](https://github.com/facebookincubator/nfusr), the Cloudserver pods would be able to (ideally) safely mount the the NFS exports. This is based on the assumption that a userspace NFS mounts doesn’t require privileges. However, this hasn’t been fully validated.
+
```
vrancurel: We also have our own NFS-client: https://github.com/scality/node-nfsc

Issues of the NFS client approach:
- Performance: We would have to implement correct connection pooling and parallelism to be credible.
- NFS compatibility: it would be hard to be compatible with all NFS servers existing, compared to using the Linux in kernel NFS client which is proven.
- We must support NFSv3 and NFSv4.

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,108 @@
...
+#### Accessing NFS
+
+The manager will update the configuration of the Cloudserver containers to mount a ReadWriteMany PV that is backed by a given NFS export. This would allow for the Cloudservers to access data as soon as it is ingested. The ingestion itself would be handled by Kubernetes CronJobs defined by the manager that runs an rclone process to completion on the specified interval. This interval can be altered or delayed by the manager as needed.
+
+#### CronJobs
+
+Utilizing rclone, each CronJob will deploy a pod that has the duty of syncing the metadata from the NFS mount to Cloudserver on the schedule provided. These jobs will run until completion and will not schedule another job until the previous has finished unless the manager determines that the job needs to be re-run earlier based on the contents of the geosync log.
+
```
vrancurel: We are using the geosync log as a hint for feeding rclone, but not a source of truth.

vrancurel: Also using the listing/attribute cache to speed up the listing.

vrancurel: Acceleration of scanning can be done in V2.

03-out-of-band-nfs.md
vrancurel: Please mention about reading/writing files and also the special header with MD5 and size.

vrancurel: Also please include timeline for V1, V2 and V3.

ssalaues: Another options is providing a gateway directly to the file system over having listings via mongo

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,108 @@
+# ZIP#3 (Zenko Improvement Proposal) Out-of-Band NFS Metadata Ingestion
```
ssalaues: Remove references to `metadata`

mastachand:
A few comments:

- I know that this is only specifically talking about NFS. What about SMB shares? Is this another design? Nagoya, who will most likely be our first customer using this, is going to be running SMB. For SOFS, indeed, we could mount the same data via NFS, but some NAS don't necessary have that option (or it's disabled by the customer). Also, NFS and SMB ACLs don't always work well together, so I think it would be better to stick to the same protocol that the customer is using.
- For SOFS only, have you thought of leveraging our CDMI (dewpoint) interface into a SOFS catalog, that keeps the out-of-band updates in an HTTP realm. One of the issues is that we need a dewpoint connector somewhere. Should that be in Zenko? not sure this would be the best, but then it requires additional resources on the RING side (VM). Worth looking into though.

- with this design, are all cloudservers mounting the shares/exports we want to fetch metadata from? or just Atmosphere?
- in this design, it also seems that the mounts are done in a RW mode. Can't this just be done RO? To avoid any changes.

vrancurel:
On the alternatives, please also mention the listing done in "real time" in the "FS" pass-through location, and different issues (MD5 computation, initial ingestion, etc)

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,150 @@
...
+
+  **Cons:**
+
+  - There is a complication when file system clients perform partial updates of the files
+
+  - Only happens with specific file system workloads, e.g. random writes. And this constraint can be easily understood (or stated as not optimized)
+
+  - The first HEAD/GET on the partially modified file the mongodb image gets updated
+
+  - Races between readdir and stat are more commonly understood, which can be ok as long as the former returns the truth about the file.
```
ploki:
```suggestion
  - Races between readdir and stat are more commonly understood, which can be ok as long as the latter returns the truth about the file.
```

ploki: In addition, this race is also possible when using rclone.

And for the case when the filesystem and the database contents are not in agreement. It has to be handled according to the file system content, no matter how the hierarchy and attributes are kept in sync.

 My intuition tells me that all the items listed in the cons are not really cons because I think all of them also apply to the "with rclone" implementation

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,150 @@
+<h1 align="center">
```
smaffulli: remove this html-ism and use proper markdown for H1
```suggestion
<h1 align="center">
```

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,150 @@
...
+ZIP#3 (Zenko Improvement Proposal)<br>The "Cosmos" Storage Backend
+</h1>
```
smaffulli:
```suggestion
```

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,150 @@
+<h1 align="center">
+ZIP#3 (Zenko Improvement Proposal)<br>The "Cosmos" Storage Backend
```
smaffulli:
```suggestion
# ZIP#3 (Zenko Improvement Proposal)<br>The "Cosmos" Storage Backend
```

03-out-of-band-nfs.md
giacomoguiulfo: Why?

giacomoguiulfo:
@mastachand In response to your comments:

- The design has changed considerably since your first review. The updated one will support SMB shares and much more.
- With this new design, out-of-band updates will be kept in an HTTP realm.
- Only the `cosmos` pods (`fs-data` & `rclone` containers) will mount to exports/shares.
- v1 will only support RO. For v3, there will be full RW support. However, this will be configurable to match the user's needs (most likely, RO will be the default).

03-out-of-band-nfs.md
smaffulli: because content and representation should be separate. The whole reason to have markdown is not to deal with things like 'align=center' and manage that in the CSS :)

giacomoguiulfo: IMHO, that separation of concerns applies more to applications or static sites (ala Jekyll or Hugo), not really `README`-like files. Without this style, the doc doesn't look as nice, and AFIAK, we don't have CSS here :)

Here is a list of [awesome-readmes](https://github.com/matiassingers/awesome-readme) that do the same:
- https://github.com/choojs/choo/blob/master/README.md
- https://github.com/anfederico/Clairvoyant
- https://github.com/amitmerchant1990/electron-markdownify

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,153 @@
...
+
+- CephFS
+
+- FlexVolume
+
+- [and more](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes)
+
+### Technical Details
+
+In short, a manager called *Atmosphere* will watch MongoDB for configuration updates. If a new storage location of type "filesystem" is added, then *Atmosphere* will provision a *Cosmos* pod and a persistent volume of the desired type (see above). This pod will then serve that data underlying the PersistentVolume over HTTP to CloudServer, similarly to how S3-Data does it. Additionally, this pod will synchronise/ingest pre-existing data and data not created through Zenko chronologically.
```
NicolasT: As a system admin, I'd rather not give the RBAC permimssions to create PVs in my K8s cluster to random services.

As discussed before, but now FTR, I think (ab)using the K8s PV functionality for this purpose is wrong and a clear layering violation.

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,153 @@
...
+These two pods will share a mount which corresponds to a PersistentVolume of the desired backend type (NFS, SMB, and so on). 
+
+**Note:** Optimisations can be made in *v2* to allow the rclone process to run sooner based on things like geosync logs for hinting and listing/attribute cache to speed up the listing.
+
+### Alternatives
+
+Currently, there are **NO** alternatives for the "Cosmos" framework. However, there are some alternatives for supporting "NFS" compatible storage systems as a backend and/or being able to ingest data from them. Here is a list of them:
+
+- **Option 1: Mount CloudServer Pods**
+  With SYS_ADMIN or similar capabilities applied to each of the CloudServer containers, they would be able to directly “mount” to an NFS which would greatly simplify configuration. However, this opens the door to potentially severe security risk because CloudServer is an externally exposed service (in fact, the only one).
```
NicolasT: I fail to see how this is different from Cosmos? As in, why can't this Cosmos Pod mount whatever it needs?

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,153 @@
...
+
+  **Pros:**
+
+  - Easy to implement
+
+  **Cons:**
+
+  - Requires special privileges
+
+  - If the NFS server fails, CloudServer will be unresponsive
```
NicolasT: Same when using a PV.

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,153 @@
...
+- **Option 1: Mount CloudServer Pods**
+  With SYS_ADMIN or similar capabilities applied to each of the CloudServer containers, they would be able to directly “mount” to an NFS which would greatly simplify configuration. However, this opens the door to potentially severe security risk because CloudServer is an externally exposed service (in fact, the only one).
+
+  **Pros:**
+
+  - Easy to implement
+
+  **Cons:**
+
+  - Requires special privileges
```
NicolasT: `Atmosphere` would require RBAC rights to create PVs and create Pods, which may be even more of an issue ;-)

03-out-of-band-nfs.md
ssalaues: This was a tough one for us because on the one hand we need to mount NFS exports in a pod but on the other hand we want to restrict the access the pod will have on the system. In order to achieve the mount in the pod our options are pretty much limited to RBAC permissions or essentially kernel level root capabilities. The RBAC seemed to be the safer option but is open to discussion.

03-out-of-band-nfs.md
ssalaues: In order for any pod to mount an NFS, `SYS_ADMIN` kernel space capabilities would need to be added to the deployment. We were trying to stay away from this level of priviledge which is why we decided against it.

03-out-of-band-nfs.md
ssalaues: The difference between the listed alternative to the `Cosmos` option is that, regardless of using a PV or mounting an NFS directly, this would give Cloudserver a single point of failure that would in turn cause a disruption in the external facing services if the NFS/PV were to fail.

03-out-of-band-nfs.md
ssalaues: You are absolutely correct. I guess the real question is would we prefer RBAC rights over `SYS_ADMIN` capabilities on a deployment and why?

03-out-of-band-nfs.md
```diff
@@ -0,0 +1,153 @@
...
+Leverage Zenko features (multi-cloud replication, metadata search, lifecycle policies, and such) on data stored in any of the following backends transparently:
+
+- NFS
+
+- SMB
+
+- CephFS
+
+- FlexVolume
+
```
vrancurel: And also file protocols like FTP/FXP without the need of a PV.

PR #7, opened 2019-02-27
Create 04-self-contained-zenko-container-image

vrancurel:
ZIP to cover the self-contained Zenko container image

### DIFF
--- a/04-self-contained-zenko-container-image.md
+++ b/04-self-contained-zenko-container-image.md
@@ -1,0 +1,60 @@
+# ZIP4 - SELF CONTAINED ZENKO CONTAINER IMAGE
+
+## Overview
+
+This ZIP is about having a very simple way to deploy Zenko without having to install Kubernetes or Minikube. Ideally
+by just doing 'docker run'.
+
+## Problem Description
+
+_Detailed explanation of the problem space of the ZIP. Use as many
+words as necessary._
+
+The original discussion in Zenko forum discussion can be found [here](https://forum.zenko.io/t/how-to-run-zenko-on-a-single-machine/504).
+
+## Use-cases Description
+
+_What uses-cases does this address? what impact on actors does this
+change have?_
+
+We need to provide a way for users to easily deploy Zenko without worrying too much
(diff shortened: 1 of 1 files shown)


04-self-contained-zenko-container-image.md
```diff
@@ -0,0 +1,60 @@
...
+* New components could be added.
+* Default platform settings of components (e.g. at Kubernetes level) could change.
+* Incompatibility of fundamental concepts between platforms (e.g. Swarm does not have the concept of ingresses).
+
+If we had to maintain a Dockerfile with a supervisord (or equivalent) we would have to 
+version it and test it to ensure that we did not break anything. The problems would then be:
+- Potentially giving more work to developers maintaining another distribution of Zenko.
+- Holding releases because the 'demo' Zenko would break.
+
+One possibility to avoid this extra-maintainance cost could be to use [KIND - Kubernetes in Docker](https://github.com/kubernetes-sigs/kind).
```
vrancurel: See also k3s which is essentially all the kubernetes dependencies wrapped up into a single 40mb binary and uses 512mb of RAM https://k3s.io

04-self-contained-zenko-container-image.md
```diff
@@ -0,0 +1,60 @@
+# ZIP4 - SELF CONTAINED ZENKO CONTAINER IMAGE
+
+## Overview
+
+This ZIP is about having a very simple way to deploy Zenko without having to install Kubernetes or Minikube. Ideally
+by just doing 'docker run'.
+
+## Problem Description
+
+_Detailed explanation of the problem space of the ZIP. Use as many
```
smaffulli: these are not needed

04-self-contained-zenko-container-image.md
smaffulli: this makes me think that we're not ready for a ZIP yet... I suggest should run experiments with KIND and k3s and pick a winner. Then we have all the elements to fill in the `Alternatives` section, too.

PR #9, opened 2019-03-01
Create 01-workflow-engine

vrancurel:


nyemade-harmon:
@vrancurel Just looked through this and it f

nyemade-harmon:
@vietnscality I like the write up. It explains exactly what the detailed plan is and how it works. That said, I think the high level summary isn't high level enough. For a high level explanation, I think we need to explain a bit about the business case. Specifically, I think we need to explain how the workflow engine could solve real data workflow problems in a general sense. Would you mind if I added a section that did that?

vrancurel:
@nyemade-harmon yes please

01-workflow-engine.md
```diff
@@ -0,0 +1,374 @@
...
+* Ballerina
+* custom runtimes
+
+Which pretty much matches our requirements.
+
+Kubeless does support 3 ways of interacting with functions:
+
+* CLI-based: Useful for debugging and troubleshooting
+* HTTP-based: Useful for synchronous function calls.
+* Kafka Topic based: Useful for asynchronous function calls -- may not be used by Backbeat since we need to get the return code.
```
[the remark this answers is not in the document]
vrancurel: @jonathan-gramain can we have only 2 topics, one for all pending tasks, the other for all the return codes ?
how do we efficiently retrieve the code for a given task on an object ?

01-workflow-engine.md
```diff
@@ -0,0 +1,374 @@
...
+Plugins that manipulate data read it from CloudServer, with location:objectID
+specified in queues metadata. The plugin’s output (if any) is streamed to an
+object in CloudServer and the new location:objectID is stored in the queue for
+the next plugin in line.
+
+#### Possible Implementation
+
+For the function framework we are going to reuse an existing
+framework. We already tried
+[Kubeless](https://forum.zenko.io/t/deploying-cloudserver-along-with-kubeless-to-do-event-based-processing/478)
```
ssalaues: Regardless of implementation, is this something we would be packaged with Zenko and install by default?

vrancurel: Yes. 

We must probably choose one by default, but it is easy I think to be compliant with most of them "serverless agnostic" with a concept of backend.

01-workflow-engine.md
```diff
@@ -0,0 +1,385 @@
...
+* Multi-Step Workflows:
+A workflow can consist of many steps, with one or many inputs.
+* Impedance Tolerance Between Plugins:
+Because all plugins can’t process information at the same time, as well as for
+resilience, every time a workflow step must execute a specific plugin, input
+*and* output queuing (for example, in a specific topic) is performed, with the
+output being the input of the next plugin, and so on.
+* Multi-Language Support:
+Because a workflow module could be written in any language (NodeJS, Python, Go,
+C, C++, etc.) we will favor a REST API for interaction between modules.
```
bennettbuchanan: 👍

01-workflow-engine.md
```diff
@@ -0,0 +1,385 @@
...
+workflow_execute(json_spec, callback);
+```
+
+This would be very easy to use in another backbeat engine as a piggyback engine.
+
+##### Graphical Workflow Edition
+
+Somne high-level requirements:
+
+* There would be a "workflow engine" pannel.
```
bennettbuchanan: The concept of a panel doesn't seem to be explained. Sounds like it's related to the workflow UI editor—like a way of defining a workflow using text input as opposed to a drag-and-drop layout?

vrancurel:
Updated the design with the description of the workflow engine operator.

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+The typical actions are:
+
+* Create a new workflow with a name
+* List workflows
+* Edit a workflow
+* Start/stop a workflow: mark the specified workflow as active
+* Pause/Resume a workflow: pause or resume an active workflow
+
+The workflow edition allows the user to create boxes representing steps of the workflow. Boxes
+have zero or more input ports and zero or more output ports. An input port is supposed to received
```
bennettbuchanan: ~received~ receive

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+```
+
+Some boxes have only output ports because their role is to generate contexts. Some other boxes
+have only input ports because they are "terminators".
+
+Here is a non-exhaustive list of boxes:
+
+* A "new data" box with 1 output port: will generate a new context every time an object version is created or deleted.
+* A "search" box with 1 output port: will generate a new context periodically by listing or searching in the object storage metadata sub-system.
+* A "function" box with 1 input port and 1 output port: which is able to call external functions, such as Fission, Knative, Azure functions, AWS Lambda functions, Google Cloud Functions. Those external function are provided a context and are able to return a new context (typically a set of tags than can be merged or replace existing tags).
```
bennettbuchanan: ~external function are provided~ external functions provide

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+        |                    |
+      +-+                    +-+
+input | |      Function      | | output
+port  +-+                    +-+ port
+        |                    |
+        +--------------------+
+```
+
+Some boxes have only output ports because their role is to generate contexts. Some other boxes
+have only input ports because they are "terminators".
```
bennettbuchanan: Sounds like there are two types of boxes with this characteristic of having only one input port: "terminators" and also "update" boxes.

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+* A "search" box with 1 output port: will generate a new context periodically by listing or searching in the object storage metadata sub-system.
+* A "function" box with 1 input port and 1 output port: which is able to call external functions, such as Fission, Knative, Azure functions, AWS Lambda functions, Google Cloud Functions. Those external function are provided a context and are able to return a new context (typically a set of tags than can be merged or replace existing tags).
+* A "tag" box with 1 input port and 1 output port which is able to simply add a tag or execute a custom script  to modify the context.
+* A "decision" box with 1 input port and 2 output ports: This box is able to execute a basic condition on the tags set (or execute a more complex script) and orient the flow towards output1 or output2.
+* An "update" box with 1 input port: this box is supposed to replace the context on the storage.
+* A "terminator" box with 1 input port: this box is supposed to simply terminate the execution without any action.
+
+Note that all boxes are supposed to alter the context in "memory", the context is to be updated in the final "update" box.
```
bennettbuchanan: How does it resume mid-flow during a system crash if the context memory is lost? Will it restart from the beginning?

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+are executed through a queue trigger (e.g. Kafka).
+
+
+### Synchronous (fast) and Asynchronous Workflows
+
+Similarly entire workflows can be executed synchronously or asynchronously.
+
+Obviously a workflow containing an asynchronous function cannot be executed synchronously.
+
+Although generally workflows are executed after an operation is inserted in the operation journal of the metadata database, so they are in fact asynchronous. Every event generated between 2 boxes is persisted in a persistent queue (Kafka).
```
bennettbuchanan: So edges between boxes can represent e.g. a kafka queue? Am I reading that correctly?

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+
+Synchronous workflows can be triggered directly from the HTTP frontend of cloudserver, every event generated between 2 boxes will be persisted in memory (for fast non invasive execution).
+
+
+### Other Restrictions in Workflows
+
+There can be only one source box per workflow. Because otherwise it would lead to impossible synchronization issues.
```
bennettbuchanan: I'd also assume there can't be cyclic relationships between boxes. 🤔

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+           |
+           |
+    +------v---+        +---------------+  workflow +-------+
+    |Management| RESTful|Workflow Engine+---------->+ Store |
+    |Agent     +------->+Operator       |  storage  +---+---+
+    +---+------+        +-+-----------+-+               
+        |                 |           |             +-------+    
+pause/  |                 |           |             |MongoDB|
+resume  v                 v           |             +-------+
+     +--+--+            +-+-------+   |restart          |Operati
```
bennettbuchanan: ~Operati~ Operation

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+               |Function        |   |Function      |
+               |Engine (premise)|   |Engine (cloud)|
+               +----------------+   +--------------+
+
+```
+
+| Component | Description |
+| ------------- | ------------- |
+| Graphical Workflow Edition  | One can define multiple workflows using this Orbit UI component. When saving workflows, it generates a JSON file describing the set of workflows the Workflow Manager can read.  |
+| Workflow Statistics | The Workflow Manager maintains and reports statistics about objects traversing each workflow the steps of each workflow. |
```
bennettbuchanan: ~traversing each workflow the steps of each workflow.~ traversing the steps of each workflow.

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+
+It has special k8s privileges to create configmaps and restart the workflow engine containers.
+
+The operator is responsible for transforming the workflow configs from the persistent storage
+to the k8s configmaps in a way that is friendly to modifications and upgrades.
+
+When a user creates a new workflow in the UI, the new workflow will be POSTed to the workflow engine operator
+and stored in the Store. Also a new list of active workflow will be generated in the workflow configmap (which is an array of workflow files). The workflow will be applied to the new traffic only.
+
+When a user modifieds a new workflow in the UI, the modified workflow will be PUT to the workfow engine operator and persisted in the Store. The old workflow(s) will be active until all old events are purged from the event queue. The new workflow will be active for new events only (for this a new consumer group will be created). A specific stopper message will be inserted in the event queue to force stop the processing of old events.
```
bennettbuchanan: ~modifieds~ modifies

workflow-engine.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Example of "external" cloud functions
+
+* auto-tag: call Azure function + cognitive service to describe a picture (In this case the object content is POSTed to Azure).
+
+
+#### Possible Implementation
+
+For the function framework we are going to reuse existing
+serverless frameworkw:
```
bennettbuchanan: ~frameworkw~ frameworks

PR #14, opened 2020-05-08
Add Utapi design document

tmacro:


utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+```
+1000000000000010 utapi.snapshots{bucket=bucket0} [ 2 200 200 0 { "ops": { "putObject": 2 } } ]
+1000000000000010 utapi.snapshots{bucket=bucket1} [ 1 100 200 0 { "ops": { "putObject": 2, "deleteObject": 1 } } ]
+```
+
+#### Repairs
+
+In a distributed system node or network failures can cause events to not be
+immediately ingested into Utapi and can cause problems for a system rely on a
+strict ordering of historic events. To guard against this Utapi implements a
```
jonathan-gramain, line 226: Do we actually need strict ordering for utapi metrics? In the sense that missing or duplicate events would likely be unacceptable but if they are in the wrong order but all there, does it matter?

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+Utapi stores four integers, using Warp10's multivariate data type, for every
+operation it ingests, `objectDelta`, `bytesDelta`, `ingress`, and `egress`.
+They reflect how an operation affects the computed metric counters and can be
+both positive and negative.
+
+For example, a upload of a 100 byte object to a new key would have the values:
+
+```json
+{
+	"objectDelta": 1,
```
miniscruff, line 131: I would like to propose using 2 spaces over a tab for JSON samples. 2 space indents are recommended by Google and Airbnb javascript guides and are easy to copy and paste in any editor or tool. Do not care either way so if its been tabs in a lot of other places then, by all means, keep it tabs.

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+system for asynchronous repairs of checkpoints and snapshots. During the
+Redis -> Warp10 transition Utapi performs some sanity checks to determine if
+the inserted metrics could have already included in a checkpoint. If so the
+repair process is started using the beginning and end of the time range of the
+inserted events as the beginning and end of the repair range. Effected
+checkpoints are fetched and recalculated to include the new events. Snapshots
+including the updated checkpoints are then fetched and recalculated. 
+
+```
+   +-------+ <--+---+ +-------+ <--+---+ +---------+
```
miniscruff, line 236: do you have a tool to make these diagrams or did you do them by hand? Just curious.

tmacro: http://asciiflow.com/

utapi/DESIGN.md
tmacro: "static" may be more appropriate here as there is no "correct" order, but it's important that it doesn't change

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+{
+  "objectDelta": -1,
+  "bytesDelta": -100,
+  "ingress": 0,
+  "egress": 0
+}
+```
+
+Each event can also have attached metadata such as the `account` , `bucket`,
+or `location` of the affected resource. These are attached as Warp10 `labels`
```
hbs, line 150: What would the 'location' label contain? Can you foresee the cardinality of each of those labels (i.e. the number of distinct values they may have)?

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+are seen the event's values are taken and summed with any previous values
+matching the same `label` value. A count of each value of the label
+`operationId` is also kept for every checkpoint.
+
+For example, the following `events` (shown in `GTS` format for brevity) if
+used with the label `bucket` will produce two checkpoints tagged with `bucket0`
+and `bucket1` respectively.
+
+```
+1000000000000000 utapi.events{operatioinId=putObject bucket=bucket0 key=obj0} [ 1 100 100 0 ]
```
hbs, line 174: If I understand correctly, the 'key' label would contain the actual object the operation was performed on? This could potentially lead to a very high cardinality which would impact performance and may cause issues down the line.

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+
+In a distributed system node or network failures can cause events to not be
+immediately ingested into Utapi and can cause problems for a system rely on a
+static ordering of historic events. To guard against this Utapi implements a
+system for asynchronous repairs of checkpoints and snapshots. During the
+Redis -> Warp10 transition Utapi performs some sanity checks to determine if
+the inserted metrics could have already included in a checkpoint. If so the
+repair process is started using the beginning and end of the time range of the
+inserted events as the beginning and end of the repair range. Effected
+checkpoints are fetched and recalculated to include the new events. Snapshots
```
hbs, line 232: So this means you will write events at timestamps already having a value? If that is the case then it is to be taken into account to assess the possible failure scenarios using datalog replication.

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+serialized and stored in a key. Based on its timestamp the event is then sharded
+into a 10 second block and its key is added to a corresponding Set.
+
+**Key Schema**
+
+Events are stored using `<prefix>:events:<event uuid>`
+Shards are stored using `<prefix>:shards:<timestamp>`
+`timestamp` is a UNIX style timestamp indicating the beginning of the shard
+
+#### Warp10
```
hbs, line 52: Just a syntactic remark, it's Warp 10 rather than Warp10.

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+
+Warp10 is used for long term storage and analysis of events in Utapi. In Warp10
+parlance each event is a `measure` consisting of a timestamp, `class`, key value
+pair `labels`, and an associated value.
```
hbs, line 56: Could you describe the labels associated with each class.

utapi/DESIGN.md
```diff
@@ -0,0 +1,294 @@
...
+are fetched. If no previous master snapshots are found a timestamp of `0` is
+used causing all checkpoints to be included. The checkpoints are then iterated
+over and summed together based upon their labels using the values from the
+previous snapshot as a base. 
+
+For example the two checkpoints from the previous example:
+
+```
+1000000000000010 utapi.checkpoints{bucket=bucket0} [ 2 200 200 0 { "ops": { "putObject": 2 } } ]
+1000000000000010 utapi.checkpoints{bucket=bucket1} [ 1 100 200 0 { "ops": { "putObject": 2, "deleteObject": 1 } } ]
```
hbs, line 212: Do you intend to store values as JSON? If so you might want to consider other less costly alternatives.

utapi/DESIGN.md
tmacro: yes keys a high number of unque keys is expected, additionally there is a uuid label thats unique per measure

utapi/DESIGN.md
tmacro: Yes, they're updated "in place"

utapi/DESIGN.md
tmacro: I've played with packing values into a multivariate type, do you have a suggestion to emulate a dict like structure?

utapi/DESIGN.md
tmacro: Based on your other comments, it sounds like labels should contain a minimum of unique values for performance reasons. Is there a preferred way to tag measure so they can be easily filtered and search?

utapi/DESIGN.md
tmacro: The POC is currently using something like
```js
{ 
  operationId: 'putObject',
  bucket: 'test00',
  key: 'imakey',
  account: '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be',
  location: 'site1',
  user: 'myuser',
  uuid: '8b660e04-028d-4329-924a-3af3ddb707f7'
}
```

utapi/DESIGN.md
hbs: Having a unique id per measure stored as a label in a time series database is a show stopper as you will then not have series but singletons.

The only things which should be put in labels are those elements which will help you organize your data, anything else should be stored as a value, either in its own series or as part of a complex value (either MultiValue or a serialized structure)

utapi/DESIGN.md
hbs: The upcoming version of Warp 10 will offer a way to interpret any value at ingestion time using a WarpScript macro. We will have a blog post explaining how this feature can be leveraged to store protobuf structures which are serialized on Warp 10 side, this would probably fit your needs by offering the most flexibility.

utapi/DESIGN.md
hbs: Can you estimate the cardinalities of those various elements (uuid set aside, see comment below)? A rough estimate, and is each one capped or can it grow boundlessly in terms of possible values.

tmacro: `operationId` will be a fixed number <100
`bucket` and `key` could be in the billions, unbounded
`account and `user` < 10,000  unbounded
`location` <100 technically unbounded but would grow very slowly

utapi/DESIGN.md
vrancurel: @hbs what are the possible failure scenarios you are talking about in this case ?
Thansk

PR #18, opened 2020-05-27
documentation: S3C-2675 add audit logs requirements document

terrydervaux:
Add audit logs requirements document for S3 components.


docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+* In case of an integration with an External Identity Provider(using web identity or SAML federation), the 'who' must be the External Identity login or username
+* Failed action must appear in the Audit logs with the reason of failed (e.g: wrong signature, or privileges...)
+* Audit logs should be included into sreport
+* Audit logs rotation, purge, activation, desactivation must be configurable
+* Audit logs configuration, activation, desactivation must be documented
+* Audit logs configuration, activation, desactivation must persist to Scality upgrade
+
+### Who definition
+
+'Who' can be:
```
RachedBenMustapha, line 26: There is the effective identity that performs an action, and the actual identity behind. For example in the case of a scheduled task, the "who" can be a service account, on its own, but for an asynchronous task submitted by a real user and performed by a service account, the "who" is the combination of:
- the real user that initiated the action
- the service account that performed the action on behalf of the user above
- the principal used (a user can have several ways to identify, like several access keys, roles to assume, etc)

Here service account is very broadly defined, can be unix-level or IAM level or database-level.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Context
+
+Some customers are using tools like [IBM QRadar](https://www.ibm.com/fr-fr/security/security-intelligence/qradar) in order to centralize and monitor who did what on IT systems for security reason. This tool is ingesting events formatted in Leefx format, which is a basic key/value list, separated with a space.
+
+## Requirements
+
+### Storie
+
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
```
RachedBenMustapha, line 11: Focusing only on S3/IAM level events/actions is an incomplete way of tracking changes to the system: actions that relate to security (and can open breaches, unauthorized access, broken functionality) can be in other layers such as system/unix, database, orchestrator, authentication subsystem.

If we want to exclude these other layers, I suggest stating it explicitly in the specification.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+
+* Audit logs must be formatted using LEEF format
+* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a ressource
+* In case of an integration with an External Identity Provider(using web identity or SAML federation), the 'who' must be the External Identity login or username
+* Failed action must appear in the Audit logs with the reason of failed (e.g: wrong signature, or privileges...)
+* Audit logs should be included into sreport
+* Audit logs rotation, purge, activation, desactivation must be configurable
+* Audit logs configuration, activation, desactivation must be documented
+* Audit logs configuration, activation, desactivation must persist to Scality upgrade
+
```
RachedBenMustapha, line 23: There are durability, immutability, resilience and performance considerations as well.

docs/requirements/AUDIT-LOGS.md
nootal: Agreed. It's only related to S3 components Scality is delivering. We don't care about logging the SSH connections here.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+## Requirements
+
+### Storie
+
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
+
+Acceptance Criteria:
+
+* Audit logs must be formatted using LEEF format
+* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a ressource
```
nootal, line 16: Careful here, because the `who` can be misleading. If you only have a key to access a resource, you may need to correlate with other sources to have a name for example.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
+# [S3C-2675](https://scality.atlassian.net/browse/S3C-2675): Audit logs requirements document
+
+## Context
+
+Some customers are using tools like [IBM QRadar](https://www.ibm.com/fr-fr/security/security-intelligence/qradar) in order to centralize and monitor who did what on IT systems for security reason. This tool is ingesting events formatted in Leefx format, which is a basic key/value list, separated with a space.
```
anurag4DSB, line 5: It would be nice to know which LEEF version are we going to transform out logs to.
The headers look something like this:
• LEEF:Version|Vendor|Product|Version|EventID|
• LEEF:1.0|Microsoft|MSExchange|4.0 SP1|15345|
• LEEF:2.0|Lancope|StealthWatch|1.0|41|^|

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Requirements
+
+### Storie
+
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
+
+Acceptance Criteria:
+
+* Audit logs must be formatted using LEEF format
```
anurag4DSB, line 15: It would be nice to know which LEEF version are we going to transform the logs.
The headers might differ based on the version. The header will look something like this:
• LEEF:Version|Vendor|Product|Version|EventID|
• LEEF:1.0|Microsoft|MSExchange|4.0 SP1|15345|
• LEEF:2.0|Lancope|StealthWatch|1.0|41|^|

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+* Audit logs should be included into sreport
+* Audit logs rotation, purge, activation, desactivation must be configurable
+* Audit logs configuration, activation, desactivation must be documented
+* Audit logs configuration, activation, desactivation must persist to Scality upgrade
+
+### Who definition
+
+'Who' can be:
+
+* Account root user
```
anurag4DSB, line 28: If we to audit actions performed on an account, like create account of modify its entities, like quota, keys, injecting account IDs etc. Wo can be the master admin as well, whose credentials are used to create accounts.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+* IAM users
+* Federated users (using web identity or SAML federation)
+
+### IAM ressources definition
+
+IAM ressources can be:
```
anurag4DSB, line 34: What about accounts?

nloewensen: I agreed that accounts need to be included, specifically, since we cannot ensure that the customer will not use his/her account-credentials to do some actions.

In addition, the action of the super-admin (account creation/deletion/modification) need to be logged as well

docs/requirements/AUDIT-LOGS.md
nloewensen: Charles and team already have a good understanding of the formatting and its options, so please check with them for details.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
+# [S3C-2675](https://scality.atlassian.net/browse/S3C-2675): Audit logs requirements document
+
+## Context
+
+Some customers are using tools like [IBM QRadar](https://www.ibm.com/fr-fr/security/security-intelligence/qradar) in order to centralize and monitor who did what on IT systems for security reason. This tool is ingesting events formatted in Leefx format, which is a basic key/value list, separated with a space.
+
```
nloewensen, line 6: I am not clear if we will provide audit-logs in LEEF format "only" or if we plan to provide them by default in another format and only on customer request the LEEF format will be enalbed as as addition? This should be fully documented, so that ideally customers can enable it themselves.

nloewensen: One addition for Amadeus: they do not require the logs in "real-time" but poll the logs every few minutes and process them about every 10min.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
+
+Acceptance Criteria:
+
+* Audit logs must be formatted using LEEF format
+* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a ressource
+* In case of an integration with an External Identity Provider(using web identity or SAML federation), the 'who' must be the External Identity login or username
+* Failed action must appear in the Audit logs with the reason of failed (e.g: wrong signature, or privileges...)
+* Audit logs should be included into sreport
```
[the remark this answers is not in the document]
scality-trevorbenson: There is a project called [SOSCleaner](https://github.com/soscleaner/soscleaner) to obfuscate sensitive information from sosreports. It is available via EPEL.

docs/requirements/AUDIT-LOGS.md
anurag4DSB: So shouldn't it be a part of requirements?

docs/requirements/AUDIT-LOGS.md
RachedBenMustapha: I will depend entirely on the implementation whether it's easy to make immutable (flat files vs something else), and whether we can rely solely on containerization for isolation or if more mechanisms are needed.

I consider durability and resilience to be as important as immutability: if part of the audit trail can be lost or unavailable when part of the system is down (permanently or temporarily) say before it has been "published" to the IBM software, then there can be a time period during which there is no audit, which tend to be the exact time when the audit trail would need to be consulted.

docs/requirements/AUDIT-LOGS.md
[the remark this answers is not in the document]
scality-trevorbenson: > assuming Rached only refers to the "include into sreport". Indeed, when including them in sreport or sending them to our Support in any other way, we need to stripped/obfuscate/replace any sensitive information (AD-names, user/account names etc.) Some customers even consider IP-addresses as sensitive

@nloewensen There is a project called SOSCleaner to obfuscate sensitive information from sosreports. It is available via EPEL.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
+# [S3C-2675](https://scality.atlassian.net/browse/S3C-2675): Audit logs requirements document
+
+## Context
+
+Some customers are using tools like [IBM QRadar](https://www.ibm.com/fr-fr/security/security-intelligence/qradar) in order to centralize and monitor who did what on IT systems for security reason. This tool is ingesting events formatted in Leefx format, which is a basic key/value list, separated with a space.
+
+## Requirements
+
+### Storie
```
NicolasT, line 9:
```suggestion
### Story
```

docs/requirements/AUDIT-LOGS.md
NicolasT, line 5: Is there some link to the spec of this *Leefx* format?

docs/requirements/AUDIT-LOGS.md
NicolasT, line 11:
```suggestion
As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order to connect my Scality products to my IBM QRadar system.
```

NicolasT: FWIW, I asked the whole architect team to spend some time reading this spec, since its essence goes way beyond S3C (and future Zenko): we should figure out how to provide a unified audit logging system throughout our software stack (indeed likely not including OS-level auditing which is outside our software's scope), not end up with a different approach per component.

docs/requirements/AUDIT-LOGS.md
NicolasT: Indeed, this ought to be defined. Furthermore, in the intro, there's a reference to *Leefx*. What's the right name?

docs/requirements/AUDIT-LOGS.md
NicolasT, line 16:
```suggestion
* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a resource
```

docs/requirements/AUDIT-LOGS.md
NicolasT: Agreed. Also, knowing (part of) an audit trail is 'missing' could be an important requirement.

docs/requirements/AUDIT-LOGS.md
NicolasT, line 34:
```suggestion
IAM resources can be:
```

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+
+IAM ressources can be:
+
+* Users
+* Policies
+* Groups
+* Identity provided by Identity provider
+* Roles
+
+## IAM Actions definition
```
NicolasT, line 42: Instead of semi-exhaustively listing these calls: are there any calls *not* listed here? If so, why? If not, shouldn't the simply requirement be "All operations are to be logged", given we may extend these APIs in the future anyway.

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
+
+Acceptance Criteria:
+
+* Audit logs must be formatted using LEEF format
+* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a ressource
+* In case of an integration with an External Identity Provider(using web identity or SAML federation), the 'who' must be the External Identity login or username
+* Failed action must appear in the Audit logs with the reason of failed (e.g: wrong signature, or privileges...)
+* Audit logs should be included into sreport
+* Audit logs rotation, purge, activation, desactivation must be configurable
```
benoit-a, line 20: Does it mean "rotation = yes/no , purge = yes / no" or something more subtle like "max MB rotation / purge" or "max 3 days", etc... ?

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+* GetBucketTagging
+* GetBucketVersioning
+* GetBucketWebsite
+* GetObject
+* GetObjectAcl
+* GetObjectLegalHold
+* GetObjectLockConfiguration
+* GetObjectRetention
+* GetObjectTagging
+* GetObjectTorrent
```
benoit-a, line 246: Do we support that ?

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+* Objects
+
+### S3 actions definition
+
+S3 actions can be:
+
+* AbortMultipartUpload
+* CompleteMultipartUpload
+* CopyObject
+* CreateBucket
```
benoit-a, line 203: Question: do we plan to log any additional to the request ? like creating a Bucket "public" or something like that ?

Same for acl below. Logging "change of acl" is not very useful if you don't know anything that was changed. (assuming the purpose of the log is to trace someone's actions)

docs/requirements/AUDIT-LOGS.md
benoit-a: I would agree to Nicolas comment.

Maybe that's just a wording but I would rephrase like:
"All IAM actions needs to be logged" (and listing exceptions if they are any) . Eventually, the list can be introduced with "including those actions (but not limited to) below"

On a moving target, a future proof definition (of what should be part of the log) should be included alongside a list.

docs/requirements/AUDIT-LOGS.md
vrancurel: Can we just consider it is a log like other logs with a marker saying it is an audit log ? Then we don't have to redesign specialized distributed log system ...

NicolasT: What do you mean exactly with 'marker'? A specific field in the log output? Then how would this be handed over to an external system (assuming no post-processing)?

docs/requirements/AUDIT-LOGS.md
```diff
@@ -0,0 +1,292 @@
...
+### Storie
+
+As a Storage Administrator, I want audit logs that describe who perform an action on a S3 or IAM ressource, in order connect my Scality products to my IBM QRadar system.
+
+Acceptance Criteria:
+
+* Audit logs must be formatted using LEEF format
+* Audit logs must be self-descriptive, that means a Storage Administrator don't need to correlate multiple log lines or log files to know who perform an action on a ressource
+* In case of an integration with an External Identity Provider(using web identity or SAML federation), the 'who' must be the External Identity login or username
+* Failed action must appear in the Audit logs with the reason of failed (e.g: wrong signature, or privileges...)
```
Stephane-Scality, line 18:
```suggestion
* Failed action must appear in the Audit logs with the reason of failure (e.g: wrong signature, or privileges...)
```

docs/requirements/AUDIT-LOGS.md
Stephane-Scality: I believe rotation itself should provide guarantees, a user most likely is required to know that there are no holes in the log, so if rotation deletes a file it must be the oldest file on this system, failure to rotate/delete the oldest file stops rotation/deletion.

docs/requirements/AUDIT-LOGS.md
Stephane-Scality: AFAIK most systems provide tampering detection rather than guaranteed tampering prevention (the guaranteed being very hard to do). It is in general sufficient to provide signatures that can be verified for a single trail, then for a group of trails, etc. in a "tree" form.

PR #20, opened 2020-07-06
feature: bucket notification design

vrancurel:
Initial commit.

### DIFF
--- a/docs/design/bucket-notification.md
+++ b/docs/design/bucket-notification.md
@@ -1,0 +1,303 @@
+# Bucket Notification
+
+The bucket notification feature enables an external application to
+receive notifications when certain events happen in a bucket.
+
+The API part of the feature will be built in line with the AWS
+specification while the notification delivery system will be specific
+to our solution, and will permit the delivery through a variety of
+mediums (e.g. RabbitMQ, Kafka, etc).
+
+We plan to leverage the
+[Backbeat](https://github.com/scality/backbeat) component to perform
+the collection of events from the S3C Raft oplog (and later on from
+MongoDB oplog), the queuing and the delivery of events.
+
+The feature has to be implemented mainly for S3C and eventually be
+implemented for Zenko and Blobserver.
+
+## Requirements
+
+## Use-cases Description
+
+## API
+
+The `PutBucketNotificationConfiguration` API will store the bucket
+configuration in a post-processed way (XML converted to JSON) into the
+`metastore` special database.
+
+The structure of a BucketNotificationConfiguration payload looks like
+the one defined in the [AWS
+spec](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html)
+but is slightly different:
+
+In the AWS spec it is possible to specify:
+
+* TopicConfiguration: for notifying an event (SNS)
+* QueueConfiguration: for queueing an event (SQS)
+* CloudFunctionConfiguration: for executing a cloud function (Lambda)
+
+We plan to only support `QueueConfiguration` for now.  For each
+`QueueConfiguration` there are 4 types of items:
+
+* `Id`: a unique user defined ID or automatically generated.
+* `Event`: the type of event.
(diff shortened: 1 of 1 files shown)


docs/design/bucket-notification.md
```diff
@@ -0,0 +1,284 @@
...
+specification as it is broad enough (we may later extend it for more
+granularity).
+
+### Backbeat Changes
+
+A specific database oplog consumer (here Metadata oplog, but later
+MongoDB oplog) will filter out the metadata operations according to
+the filters defined in the bucket configuration, interpret the
+`originOp` field and queue the operations in a specific Backbeat topic
+`bucket-notification-topic`.
```
jonathan-gramain, line 152: We can probably remove the `-topic` suffix from the topic name

docs/design/bucket-notification.md
```diff
@@ -0,0 +1,284 @@
...
+        <QueueArn>arn:scality:bucketnotif:::target2</QueueArn>
+    </QueueConfiguration>
+</NotificationConfiguration>
+```
+
+Event types that will need to support:
+
+Event | Description
+------|------------
+`s3:ObjectCreated:*`, `s3:ObjectCreated:Put`, `s3:ObjectCreated:Post`, `s3:ObjectCreated:Copy`, `s3:ObjectCreated:CompleteMultipartUpload` | Will generate an event respectively every time an object is created (whatever the operation type), written, created, copied, MPU constituted.
```
jonathan-gramain, line 81: Could differentiate in the description the "created (whatever the operation type)" from the "created", with a bit more wording to precise what is meant (I'm not sure myself). Same thing below between "deleted (whatever the operation type)" and "deleted", where I guess that the latter means "delete a specific version"?

docs/design/bucket-notification.md
```diff
@@ -0,0 +1,284 @@
...
+
+Note: For S3C listening to all the `metastore` oplog events is
+scalable because it has its own raft session and we don't expect to
+have too many entries. In the case of Zenko, we may have to do an
+initial discovery directly into the `metastore` bucket, and then the
+oplog is rotated so we expect it to be relativelty fast to process.
+
+For each NotificationConfiguration detected we will update a specific
+Zookeeper entry in the path
+`/bucket-notification`. E.g. `/bucket-notification/<BucketName>/MyFilterSet1`
```
jonathan-gramain, line 204: It could be good enough, but we should compare against the other paths used to keep some consistency as I can remember we use `/com/scality/backbeat` prefix e.g. for replication, not sure about ingestion, it could be overkill but just to make sure we organize the paths cleanly, especially if at some point we might be going to share a zookeeper cluster with RING services (or next-gen).

docs/design/bucket-notification.md
```diff
@@ -0,0 +1,284 @@
...
+where `BucketName` is the name of the bucket, `MyFilterSet` is the
+unique id defined in the `QueueConfiguration`. The file will contain
+all the filter rules and the target in JSON format.
+
+#### Bucket Notification Queue Populator: Event Processing
+
+Each time the Bucket notification queue populator will match an event
+in the oplog according to the active set of filters, it will create an
+entry in a the Bucket Notification topic. For sake of disk space we
```
jonathan-gramain, line 213: ~in a the~ in the

docs/design/bucket-notification.md
```diff
@@ -0,0 +1,284 @@
...
+
+#### Bucket Notification Processor: Actions
+
+The bucket notification processor will adapt dynamically to bucket
+notification configuration changes (added / deleted filter rules) by
+setting a watcher on the bucket notification Zookeeper path.
+
+Each target will have its own processor on the topic, consume at its
+own pace, skip entries which do not match and produce in its external
+endpoint. Parallelism will be achieved by having consumer groups (one
```
jonathan-gramain, line 231: Not sure I understand the meaning of "one consumer per partition" here, if you have consumer groups, you end up with multiple consumers per partition in the end, right? Inside a single consumer group, indeed only one consumer will consume a single partition.

PR #23, opened 2020-08-04
ft: ZENKO-2712 tape design

vrancurel:
Baseline of the technical spec.


vrancurel:
Rather than `tape` maybe call it `cold-storage` or `archive`.

docs/design/tape.md
```diff
@@ -0,0 +1,244 @@
...
+        "details": {
+           "glacier": {
+              "Vault": "my-vault"
+           }
+        }
+    },
+...
+```
+
+Locations marked as `isArchive=true` will have the following
```
jonathan-gramain, line 109: Maybe using the `type` attribute value is enough to know if it's an archive type or not internally?

The reason is I don't see if there's a use case e.g. for type="aws" and isArchive=true, or conversely, type="glacier" and isArchive=false, so it might be more of an internal strong property of the location type rather than a customizable location property, I may be wrong.

docs/design/tape.md
```diff
@@ -0,0 +1,244 @@
...
+When putting an object in an archive location, the object status will
+be marked as `ArchivePending` and its content will in fact be uploaded
+in a hidden bucket. This bucket will have special storage location
+called `tape-archive` which could be mapped to `file` or a RING or any
+other temporary storage.
+
+Those special buckets will have the following names
+`__tape-archive-<storage-location>-` suffixed by a random string and
+will be listable/accessible only by the root account.  There will be
+only X current archive buckets (e.g. 1).
```
jonathan-gramain, line 130: We probably want at least 2 buckets, so that when one is closed, the other can still be used to absorb new objects (a form of double buffering). Wondering also if it would help to force always having at least one open tape bucket at any time to avoid stalling archival requests, hence with 2 buckets having a form of double buffering. Just a random thought.

docs/design/tape.md
```diff
@@ -0,0 +1,244 @@
...
+a `tape-restore-object` topic. A specific `tape-restore` backbeat
+extension will then fetch the object from the configured backend and
+will restore it in a special bucket.
+
+Those special buckets will have the following names
+`__tape-restore-<storage-location>-` suffixed by a random string and
+will be listable/accessible only by the root account.
+
+The backbeat task will then copy the objects from this specific
+storage location to their initial storage location. Update the object
```
jonathan-gramain, line 168: typo: ~location. Update~ location, update...

docs/design/tape.md
vrancurel: Yes it seems a good idea.

docs/design/tape.md
vrancurel: Yes, indeed, it shall be an internal property of the storage location.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,384 @@
...
+...
+```
+
+If the object restore is in progress.
+
+### Cloudserver Changes
+
+* Mark objects are ArchivePending, or RestorePending.
+
+* A special behind-the-scene non-intrusive S3 PUT operation to restore
```
vrancurel, line 346: Potentially supporting MPU for large objects.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,381 @@
...
+
+We need to manage the special `x-amz-restore` header to indicate the
+status of the retrieval.
+
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html
+
+## Technical Details
+
+### Location Config
+
```
pgueant, line 158: A bucket must be able to list all of its objects, regardless of the storage class each object belongs to.
Storage locations should be mapped to data centers or availability zones, not to storage classes. Storage locations are different from storage classes.  The tape backend should not be seen as a location but as a storage class.
The concept of subLocations is unnecessarily complex. Different replication configurations should be mapped to different storage classes.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,381 @@
...
+* Impossibility to create a bucket on those locations
+* Objects can be created/lifecycled/restored
+
+The `fallback` location will be used for direct PUTs in the cold
+storage location. While waiting to the real archiving to kick in.
+
+The `subLocation` is used to specify the replication level or the type
+of device used (in case of multiple choices available - this is cold
+storage backend dependent).
+
```
pgueant, line 204: These really should be new storage classes, e.g. storage class tape with 1 replica or storage class tape with 3 replicas

pgueant (changes requested):

docs/design/cold-storage.md
vrancurel: These are some parameters we directly pass to the tape library.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+archive constitution will be specific to our solution, and will permit
+the delivery through a variety of mediums (e.g. Glacier, DMF,
+StrongLink, SpectraLogic, etc).
+
+We plan to leverage the
+[Backbeat](https://github.com/scality/backbeat) component to perform
+the upload and restore of objects in batches and maintain the object
+statuses.
+
+The feature has to be implemented mainly for XDM.
```
jonathan-gramain, line 18: and/or Artesca?

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+| Component  | Description |
+| ------------- | ------------- |
+| application | Any application that uses S3 Glacier API |
+| zenko_xdm | The component that we describe in this spec |
+| glacier | A cloud based cold storage service |
+| tape_library | A tape library |
+
+### Category 2
+
+In this category data is transfered through a POSIX FilesyStem,
```
jonathan-gramain, line 108: typo: File**S**ystem (or filesystem)

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+exposed by the gateway. Archived data is indexed through stubs in the
+filesystem.
+
+The problems of this category are:
```
jonathan-gramain, line 112: Maybe also explain the benefits, including possible reasons why there are no better alternatives (is it because the tape library only talks FS?)

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+* It has the same problems as category 2. Especially lack of control on the archival/restore process.
+
+<!--
+@startuml img/cold-storage/overview-category-3.png
+skinparam BackgroundColor transparent
+allow_mixing
+
+agent application
+agent zenko_xdm
+agent s3_interface
```
jonathan-gramain, line 168: Should it be called "s3_gateway" as a gateway to a filesystem?

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+@enduml
+-->
+
+![Cold storage overview Category 4](img/cold-storage/overview-category-4.png)
+
+| Component  | Description |
+| ------------- | ------------- |
+| application | Any application that uses S3 Glacier API |
+| zenko_xdm | The component that we describe in this spec |
+| s3_interface | An S3 compatible interface on top of a file system |
```
jonathan-gramain, line 233: Is it on top of a filesystem in this case?

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+`cold-storage-archive` topic, while updating its metadata with the
+`ArchivePending` status.
+
+A specific Backbeat extension consumer for each type of supported
+backend (matched by location) will monitor the archive topic and watch
+for limits to be reached.
+
+#### Limits
+
+Entries in the archive topic are checked for the following limits:
```
jonathan-gramain, line 405: It's not clear to me how those checks are done, could be worth giving some extra details.

For reference, Lifecycle today uses regular listing on a cron-style schedule to check which objects to transition to other storage classes and send messages to a kafka topic to transition them.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,626 @@
...
+cloudserver -> metadata : 3. register new object in bucket
+metadata -> backbeat : 4. notify
+backbeat -> kafka : 5. write an event in the cold-storage topic
+kafka -> backbeat_tape_library : 6. notify
+backbeat_tape_library -> tape_library : 7. Establish WebSocket (bi-directional)
+backbeat_tape_library -> tape_library : 8. Driver API "put_s3" (returns job ID)
+backbeat_tape_library -> redis : 9. Persist job ID
+tape_library -> cloudserver : 10. fetch object via S3
+cloudserver -> ring : 11. fetch object
+tape_library -> tape : 12. write object to tape
```
jonathan-gramain, line 505: In the diagram, the "tape" actor could go to the far right since it is only visible to the tape library (just for clarity). Similarly kafka could be moved between backbeat and backbeat_tape_library.

docs/design/cold-storage.md
vrancurel: There are no real benefits from an XDM point of view.

docs/design/cold-storage.md
vrancurel: In XDM we mapped the concept of storage classes to storage locations. But we can have objects with multiple storage locations in the same bucket.

docs/design/cold-storage.md
jonathan-gramain: Ok, so in other words, maybe explain why we need to implement it? I guess the reason is the tape library API support not allowing Category 1, but it's not clear to me.

docs/design/cold-storage.md
pgueant: See previous comment:
buckets have storage location
objects have storage class

This is what we sold to CNES: S3API-level compatibility to configure lifecycle transition rules between storage classes, including the cold class.
The current way XDM handles storage classes, storage locations, and lifecycle transitions diverges from AWS's specification and therefore is not suitable for how CNES projects to manage cold storage. Objects with multiple storage locations in the same bucket also diverges from AWS's specification and is not suitable to the CNES project.

docs/design/cold-storage.md
pgueant: Protection levels on the tape library should map to storage classes. The concept of "sublocation" is not present in the AWS specification and therefore should not be exposed.

docs/design/cold-storage.md
pgueant: Yes

docs/design/cold-storage.md
pgueant: See previous comment:
buckets have storage location
objects have storage class

This is what we aim for: S3API-level compatibility to configure lifecycle transition rules between storage classes, including the cold class.
The current way XDM handles storage classes, storage locations, and lifecycle transitions diverges from AWS's specification and therefore is not suitable for how customers will want to manage cold storage. Objects with multiple storage locations in the same bucket also diverges from AWS's specification and is not suitable for customers looking for S3-API compatibility.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,770 @@
...
+kafka -> backbeat_consumer : notify
+backbeat_consumer -> tape_library : Driver API "get_s3" w/ archive ID
+tape_library -> tape : read object from tape
+tape_library -> cloudserver : put object via S3
+hnote over tape_library : potentially long
+tape_library -[dotted]-> backbeat_consumer : "reponse_get_s3": status
+cloudserver -> xcore : write object w/ internal version
+cloudserver -> metadata : mark object as `Restored` (set `restoreDate/Days`)
+hnote over backbeat_consumer : wait RestoreDays
+backbeat_consumer -> metadata : scan objects to delete (piggy back on existing lifecycle mechanism)
```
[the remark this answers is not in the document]
vrancurel: actually it is transition.

docs/design/cold-storage.md
```diff
@@ -0,0 +1,770 @@
...
+Description of a get request and response:
+
+```
+{
+	"method": "get_s3",
+	"data": {
+		"url": "https://test.scality.com/bucket1/object1?versionId=qtJlDTD",
+		"accessKey": "accessKey1",
+		"secretKey": "verySecretKey1",
+		"parameters": "LTOx2",
```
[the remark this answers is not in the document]
vrancurel: It seems a bit complex for a v1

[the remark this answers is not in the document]
vrancurel: There is no difficulty going to STS once we have a working product.
We are not sure their client support STS.

docs/design/cold-storage.md
terrydervaux: I agree with @pguant.

PR #24, opened 2020-08-25
feature: S3C-3157 S3Browser redesign for bucket notification

terrydervaux:
This describes the changes required to configure bucket notification feature on the S3 Browser.


docs/design/bucket-notification.md
```diff
@@ -256,7 +256,75 @@ TBD
...
+
+### Object Lock Tab
+
+![Notification Tab](img/bucket-notification/modal-bucket-view-info-tab-object-lock.png)
+
+Acceptance Criteria:
+*   The Notification tab shall contains the old **Object Lock Default Settings**
+  section.
+*   The old **Object Lock Default Settings** section name shall be replace by
+  **Default Settings**.
```
mastachand, line 290:
```suggestion
Acceptance Criteria:
*   The Notification tab shall contain the old **Object Lock Default Settings**
  section.
*   The old **Object Lock Default Settings** section name shall be replaced by
  **Default Settings**.
```

docs/design/bucket-notification.md
```diff
@@ -256,7 +256,75 @@ TBD
...
+   and `s3:ObjectRemoved:DeleteMarkerCreated`, `s3:ObjectCreated:Copy`, and
+  `s3:ObjectCreated:CompleteMultipartUpload`events can't be checked.
+*   Clicking **ADD NOTIFICATION RULE** and **EDIT NOTIFICATION RULE** shall
+  update the bucket notification configuration according to inputs using [PutBucketNotificationConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html) action.
+
+### Alternatives
+
+We could have leverage tab on the main pages to avoid modal, and provide
+a better UX. But, the development cost of that change is not compatible with
+our deadlines, and the S3 Browser will be replaced by XDM UI offline in 2021.
```
mastachand, line 327:
```suggestion
We could have leveraged tabs on the main pages to avoid modal, and provide
a better UX. But, the development cost of that change is not compatible with
our deadlines, and the S3 Browser will be replaced by XDM UI offline in 2021.
```

docs/design/bucket-notification.md
```diff
@@ -256,7 +256,75 @@ TBD
...
+*   Pressing the tab key must switch the focus between fields.
+*   If `s3:ObjectCreated:*` event is selected, then `s3:ObjectCreated:Put`,
+  `s3:ObjectCreated:Post`, `s3:ObjectCreated:Copy`, and
+  `s3:ObjectCreated:CompleteMultipartUpload` events can't be checked.
+*   If `s3:ObjectRemoved:*` event is selected, then `s3:ObjectRemoved:Delete`,
+   and `s3:ObjectRemoved:DeleteMarkerCreated`, `s3:ObjectCreated:Copy`, and
+  `s3:ObjectCreated:CompleteMultipartUpload`events can't be checked.
+*   Clicking **ADD NOTIFICATION RULE** and **EDIT NOTIFICATION RULE** shall
+  update the bucket notification configuration according to inputs using [PutBucketNotificationConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html) action.
+
```
mastachand, line 322: can we list the available ARNs in a drop-down list instead of asking to type it all? This is prone to typos that will lead to errors easily avoidable with a drop-down list.
Those are pretty static and configured at federation level so we should be able to get the list in there. 
At the very least, we should just ask for the "target" names that are in group_vars/all, and we should build the arn since it's pretty much all static besides that target name, it's always going to be:
"arn:scality: bucketnotif:region:account-id:"+target_name

docs/design/bucket-notification.md
```diff
@@ -256,7 +256,75 @@ TBD
 
 ## UI
```
mastachand, line 258: I guess i had not thought about this, but it would have been good to have the ability to create the queues in the UI (maybe the console and not the browser) instead of having to run federation to add a target. Too late for this one, but it would be good to add this in the next version

mastachand (dismissed):
This looks good if we can fix the typos and streamline the ARN input to avoid end-user typos and frustration.

docs/design/bucket-notification.md
terrydervaux: I agree with the downlist idea. But, speaking under @naren-scality control, we can't retrieve the list the available ARN through an API today, and we can't deliver that for the next release. 

For the 2nd option, let me double-check what we can do to improve the behavior.

terrydervaux: I need to be CC at the end of the week, so it's too late. I added this item in the List of Improvements for the next release.

docs/design/bucket-notification.md
terrydervaux: I added this item in the List of Improvements for the next release.

terrydervaux:
@mastachand: Thanks for your review. I fixed the typos but unfortunately, we will not be able to streamline the ARN on the first release.

docs/design/bucket-notification.md
naren-scality: @terrydervaux please bring this up in our priority meeting, we can discuss what can be done about fetching ARNs. I agree that, given our setup, we shouldn't expect users to type in ARNs (because it is customized to fit our needs).

naren-scality (approved):

rahulreddy (approved):

rahulreddy:
@terrydervaux I asked @mastachand  to review today. You will not be able to merge w/o his review removed as blocker.

PR #32, opened 2021-01-14
document release notes by pitching changie as a tool

miniscruff:
* Pitching my side project as a means of organizing our changelogs and release notes
* Generated changelog from changie https://github.com/miniscruff/changie/blob/main/CHANGELOG.md
* Generated release notes for a github release https://github.com/miniscruff/changie/releases/tag/v0.4.0

Not a very long pitch so feel free to ask questions or concerns

### DIFF
--- a/docs/release/disconnecting.md
+++ b/docs/release/disconnecting.md
@@ -34,7 +34,7 @@ The purpose of a project repository is to:
 * Create a build chain that outputs the desired installation code and packages.
 * Trigger integration tests that ensure all selected components interoperate effectively.
 * Document the project's services, and how to build, test, configure, run, monitor, and contribute to the project.
-* Offer [release notes](./release-notes.md) that document changes in the product's function and features.
+* Offer [release notes](./release-notes.md) that document changes in the project's function and features.
 
 Project releases should be kept as small as possible.
 For example, bumping the version of a component might unlock features or bugfixes.
@@ -54,7 +54,7 @@ The purpose of a product repository is to:
 * Create a build chain that outputs the desired installation code and packages.
 * Trigger integration tests that ensure all selected projects interoperate effectively.
 * Document the product's services; how to install, configure, run, and monitor the product; and how to solve product issues.
-* Offer [release notes](./release-notes.md) that document changes in the project's function and features.
+* Offer release notes that document changes in the product's function and features.
 
 Product releases are driven by customer needs and are not defined within the object squad. 
 The object squad, however, is responsible for making project releases available to the product and for helping integrate the projects into the product. 
--- a/docs/release/disconnecting.md
+++ b/docs/release/disconnecting.md
@@ -14,7 +14,7 @@ Each component must contain:
 
 * Unit and functional tests to ensure that the component works.
 * Documentation explaining the services the component provides, and how to contribute, build, test, configure, run, use, and (potentially) monitor the component.
-* Release notes
+* [Release notes](./release-notes.md)
 * Build chain for building, testing, and publishing packages
 
 Each component will thus have its own release cycle that is independent of the release cycle of the product itself.
@@ -34,7 +34,7 @@ The purpose of a project repository is to:
 * Create a build chain that outputs the desired installation code and packages.
 * Trigger integration tests that ensure all selected components interoperate effectively.
 * Document the project's services, and how to build, test, configure, run, monitor, and contribute to the project.
-* Offer release notes that document changes in the product's function and features.
+* Offer [release notes](./release-notes.md) that document changes in the product's function and features.
 
 Project releases should be kept as small as possible.
 For example, bumping the version of a component might unlock features or bugfixes.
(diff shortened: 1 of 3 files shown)


docs/release/disconnecting.md
```diff
@@ -54,7 +54,7 @@ The purpose of a product repository is to:
 * Create a build chain that outputs the desired installation code and packages.
 * Trigger integration tests that ensure all selected projects interoperate effectively.
 * Document the product's services; how to install, configure, run, and monitor the product; and how to solve product issues.
-* Offer release notes that document changes in the project's function and features.
+* Offer [release notes](./release-notes.md) that document changes in the project's function and features.
```
tcarmet, line 57: As much as I would love to see this change on the product level, Citadel may not be the place to request it as it's mainly for the object squad. At least for now :sweat_smile: 

So maybe I made a mistake of adding this release note line on the product repository.

tcarmet, line 57: also this mistake is on me
```suggestion
* Offer [release notes](./release-notes.md) that document changes in the products's function and features.
```

docs/release/disconnecting.md
```diff
@@ -34,7 +34,7 @@ The purpose of a project repository is to:
 * Create a build chain that outputs the desired installation code and packages.
 * Trigger integration tests that ensure all selected components interoperate effectively.
 * Document the project's services, and how to build, test, configure, run, monitor, and contribute to the project.
-* Offer release notes that document changes in the product's function and features.
+* Offer [release notes](./release-notes.md) that document changes in the product's function and features.
```
tcarmet, line 37: Correcting my own mistakes :sweat_smile: 

```suggestion
* Offer [release notes](./release-notes.md) that document changes in the project's function and features.
```

tcarmet (commented):
This already looks good to me, as I already went back and forth with changie and think that we would really benefit from it, and it's very worth a shot. It is also safe to use, meaning that it doesn't lock ourself to the tool itself, it will create a bunch of changelog files which a very welcome in our repositories anyway.

Couple of notes I want to add:

1. I don't know if this should be added now, or in the future, but I remember us talking offline/online about how to use this tool along with the GitWaterflow. 
So Gitwaterflow is one example, but explaining how we would deal with X scenarios like that, and giving the knowledge to it, would be nice.
I don't think we should write it now, and rather keep in mind that the step here is to approve the usage of the tool. And then we write how we handle the corner cases. that's fine by me.

2. I would like to define how components or projects should use `changie` meaning, we have sphinx documentation on projects, and markdown or txt or no documentation on some components. And as far as I understood, changie can handle both `.md` and `.rst` output (or any output). 
So should we write about how we make the implementation on components versus projects? Or is it up to the repository itself to decide and as long as it use changie we're happy?

3. I would love some feedback here from @gdemonet and @NicolasT

miniscruff:
> Couple of notes I want to add:
> 
>     1. I don't know if this should be added now, or in the future, but I remember us talking offline/online about how to use this tool along with the GitWaterflow.
>        So Gitwaterflow is one example, but explaining how we would deal with X scenarios like that, and giving the knowledge to it, would be nice.
>        I don't think we should write it now, and rather keep in mind that the step here is to approve the usage of the tool. And then we write how we handle the corner cases. that's fine by me.
> 

I would say discussing known potential issues as soon as possible is the way to go. It would be much better to rule out tools or ideas that fail some edge cases earlier rather than later. With that said, the solution we came up with was to track changes along with features as we would a single branch. But when we release we do all the changie commands per release branch and not forward port them ( or revert the forward port however that works ).

>     2. I would like to define how components or projects should use `changie` meaning, we have sphinx documentation on projects, and markdown or txt or no documentation on some components. And as far as I understood, changie can handle both `.md` and `.rst` output (or any output).
>        So should we write about how we make the implementation on components versus projects? Or is it up to the repository itself to decide and as long as it use changie we're happy?

This pitch is for using changie of course, but if any changelog tool fits our use case better then, of course, we should use it. I mention changie because I made it and I think it fits our use case the best, but I am of course biased on this point. So before I go into specifying how to use change, we should first discuss if we need a tool to solve this problem and if changie should be that tool. If this pitch is successful then a common or starting config would be my next PR.

Also yes, changie can handle any format as it deals with file paths and go templates and not a specific format. ( I would note that I only have used markdown so some tweaks may be required to fit other formats properly )

docs/release/disconnecting.md
miniscruff:
```suggestion
* Offer release notes that document changes in the product's function and features.
```

miniscruff:
Going to do a little tagging to see if anyone can take a look during kick off week, @rahulreddy @terrydervaux @wabernatScality @tmacro

NicolasT:
FWIW: within the MetalK8s team we defined our vision on how to handle changelogs and release notes in https://docs.google.com/document/d/1NUUShY9VBdLa3s0ePR5swptBtszXwM9qvFsF52L22L4/edit?usp=sharing and validated this with various people across the organization, including TechPubs. @gdemonet started a PoC (offline) and is now developing an implementation at https://github.com/scality/changelog-binder

The core premise of the system is not keeping changelog entries in separate files (which is a must for various reasons, as explained in the doc), but the way it works (well) with GitWaterFlow, treating the VCS as a database and not only looking at trees. Furthermore, we stick to using RST to ensure only a single markup language is used across changelog/release notes and the documentation, as well as because RST defines a standardized preamble format which we use to track metadata of snippets. Finally, the Python implementation allows to (re)use the library in a Sphinx plugin so we can easily embed changelog/release notes in a (rendered) Sphinx project, the framework our documentation trees are using.

miniscruff:
Closing due to staleness and design decisions. Changelog binder project was created and the decision seems to be to use it when possible. Changie has been updated a few times but without using Git as its database does not align with @NicolasT 's suggestion. The comments about using RST and Sphinx were not an issue as Changie can output RST to be included in a Sphinx rendered HTML site.

PR #37, opened 2021-02-26
feature: describe monitoring

rahulreddy:
WIP


RachedBenMustapha (approved):
I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).

rahulreddy:
> I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).

Good idea and yes that's my intention. We are going to persist here all the monitoring we are going to expose per component. For example, for monitoring CRR we would list the metrics to be exposed - populator states for raft sessions, consumer stats etc.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
+# Monitoring in S3 Connector and Zenko
+
+Both S3 Connector and Zenko deploy a plethora of services that need to be healtchecked and monitored
```
miniscruff, line 3: You mention healtchecks here but do not include a section for it.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+
+Each service built by Scality (Cloudserver, Vault, Metadata etc) as well as third-party services (Redis, Kafka, Zookeeper etc.) should be configured for metrics collection.
+Metrics can be collected by Prometheus in either of the following ways
+
+### Pushing metrics
+
+Services that don't have a persistent storage for metrics can push metrics as they occur to Prometheus' push gateway.
+
+**Note:** This has a potential of negative impact if it's implementented in the critical path of processing a request by a service when there are issues contacting the push gateway.
+Other best practices are outlined here https://prometheus.io/docs/practices/pushing/
```
miniscruff, line 34: mkdocs does not auto-link URLs that GitHub does ( `mkdocs serve` )

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Exporters
+
+Services that have metrics already stored in some form can expose the metrics via a `/_/metrics` route. The following doc https://prometheus.io/docs/instrumenting/writing_exporters/ can be used as for guidelines and best practices for writing exporters.
```
miniscruff, line 38: Same thing here about the URL

miniscruff, line 38: > Services that have metrics already stored in some form can expose the metrics

The wording of this makes it feel optional and a "nice-to-have" when I think it should be worded as a requirement.
Like "Services that can expose an HTTP route must provide prometheus metrics at `/_/metrics`."

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+to different destinations such as email, chat applications and on call notifcation systems.
+
+## Metrics Collection
+
+Each service built by Scality (Cloudserver, Vault, Metadata etc) as well as third-party services (Redis, Kafka, Zookeeper etc.) should be configured for metrics collection.
+Metrics can be collected by Prometheus in either of the following ways
+
+### Pushing metrics
+
+Services that don't have a persistent storage for metrics can push metrics as they occur to Prometheus' push gateway.
```
miniscruff, line 31: Is the reason for push metrics due to not having persistent storage? I thought it was a persistent HTTP route or similar direct access. I may be wrong here.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+In the context of S3C and Zenko, Prometheus will be used as primary source for metrics.
+
+### Alert Manager
+
+Alert Manager can be used to push alerts by setting some alerting rules in Prometheus server. The Alert Manager can push alerts
+to different destinations such as email, chat applications and on call notifcation systems.
+
+## Metrics Collection
+
+Each service built by Scality (Cloudserver, Vault, Metadata etc) as well as third-party services (Redis, Kafka, Zookeeper etc.) should be configured for metrics collection.
```
miniscruff, line 26: I think the term 'should' should be 'must' instead.

miniscruff (commented):
I think some of the terms leave a bit much open to interpretation or opt-in were more direct do this or that would be better for this type of document.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+
+Both S3 Connector and Zenko deploy a plethora of services that need to be healtchecked and monitored
+and both products have a common architecture, Prometheus, Grafana and Alert Manager to establish a
+monitoring pipeline.
+
+## Monitoring toolkit
+
+### Prometheus
+
+A stack/toolkit containing a distributed time series databases that collects metrics from jobs or exporters and recompiles
```
NicolasT, line 11: Prometheus' TSDB isn't distributed.

docs/design/monitoring.md
NicolasT, line 31: Pushgateway is not about 'persistent storage': services should keep their metrics in-memory anyway, and report them when Prometheus scrapes them.

Pushgateway is about services that are so short-living their metrics can't be scraped (i.e., their lifetime is below the scrape frequency).

In general, I don't think we deploy anything at all which requires Pushgateway at this point.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+## Metrics Collection
+
+Each service built by Scality (Cloudserver, Vault, Metadata etc) as well as third-party services (Redis, Kafka, Zookeeper etc.) should be configured for metrics collection.
+Metrics can be collected by Prometheus in either of the following ways
+
+### Pushing metrics
+
+Services that don't have a persistent storage for metrics can push metrics as they occur to Prometheus' push gateway.
+
+**Note:** This has a potential of negative impact if it's implementented in the critical path of processing a request by a service when there are issues contacting the push gateway.
```
NicolasT, line 33: Don't push :wink:

miniscruff:
> > I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).
> 
> Good idea and yes that's my intention. We are going to persist here all the monitoring we are going to expose per component. For example, for monitoring CRR we would list the metrics to be exposed - populator states for raft sessions, consumer stats etc.

I would propose including monitoring specifics as part of the component documentation ( preferably auto-generated from code somehow). Keeping Citadel in line with components with something that specific will probably be very difficult.

Creating a baseline default set of metrics we prefer or require is good though, I agree with that.

docs/design/monitoring.md
NicolasT, line 38: Metrics aren't necessarily 'stored' in some way. Also, most services should *not* use an 'exporter' but directly expose relevant metrics from inside the process directly. This should be the primary means of metrics exposure and collection.

Exporters are *only* to be used when you *can't* have in-process metrics endpoints (closed-source app,...) but you can somehow query metrics from the service (instance), e.g. by reading some `/run` file, performing a (non-Prometheus-compatible) HTTP request to the service, a JMX RPC call,... In that case, you *can* have an _exporter_ which runs (as a separate process) next to the service to be monitored (as a sidecar), exposes a HTTP endpoint which gets scraped by Prometheus, at which point this sidecar retrieves the metrics from the actual service (using whichever means), formats them in-memory into a Prometheus-compatible metrics report, and returns them as the HTTP response.

NicolasT:
> > > I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).
> > 
> > 
> > Good idea and yes that's my intention. We are going to persist here all the monitoring we are going to expose per component. For example, for monitoring CRR we would list the metrics to be exposed - populator states for raft sessions, consumer stats etc.
> 
> I would propose including monitoring specifics as part of the component documentation ( preferably auto-generated from code somehow). Keeping Citadel in line with components with something that specific will probably be very difficult.
> 
> Creating a baseline default set of metrics we prefer or require is good though, I agree with that.

The Prometheus metrics format is self-descriptive: it has type information, and a human-readable 'documentation' string with every metric.  So all that's needed should be there when you scrape.

NicolasT:
> I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).

Also, make sure these are (where applicable) in line with 'standard' metrics used in the Prometheus ecosystem, with relevant labels etc. Think indeed, runtime resources, HTTP request metrics, queue metrics,...

miniscruff:
> > > > I don't know if this is planned yet, but later a description of standard metrics we expect all services to export would be good (like runtime resources, requests/messages processed, ...).
> > > 
> > > 
> > > Good idea and yes that's my intention. We are going to persist here all the monitoring we are going to expose per component. For example, for monitoring CRR we would list the metrics to be exposed - populator states for raft sessions, consumer stats etc.
> > 
> > 
> > I would propose including monitoring specifics as part of the component documentation ( preferably auto-generated from code somehow). Keeping Citadel in line with components with something that specific will probably be very difficult.
> > Creating a baseline default set of metrics we prefer or require is good though, I agree with that.
> 
> The Prometheus metrics format is self-descriptive: it has type information, and a human-readable 'documentation' string with every metric. So all that's needed should be there when you scrape.

True but to be able to scrape does require you to run the service first, I was thinking it would be nice to include some information as part of a markdown sort of file in the source code or docs. Maybe the prom-client has a "sample" export we can generate... But yes good point.

docs/design/monitoring.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Grafana
+
+Grafana is a visualization tool that can have many data sources such as Prometheus, custom data exporters, Elastic Search etc.
+In the context of S3C and Zenko, Prometheus will be used as primary source for metrics.
+
+### Alert Manager
+
+Alert Manager can be used to push alerts by setting some alerting rules in Prometheus server. The Alert Manager can push alerts
+to different destinations such as email, chat applications and on call notifcation systems.
```
jonathan-gramain, line 22: typo: notif**i**cation

docs/design/monitoring.md
rahulreddy: True, I will remove it. I was thinking of the storage integrations they have but it's not relevant to our stack, at least now.

docs/design/monitoring.md
rahulreddy: I think our reasons and design changed that the original usecase no longer exists. I will remove the push gateway section.

docs/design/monitoring.md
rahulreddy: I was using the word `exporter` interchangably with a service exposing metrics on its on endpoint. I understand the difference between the two but I see that's how its used in Prometheus documentation as well.

docs/design/monitoring.md
rahulreddy: Ah, now I remember. Pushgateway was being suggested for background jobs or cronjobs which are not services responding to HTTP requests. The suggestion is to daemonize these cronjobs which most of them already, so they run as a service on a port exposing healthcheck and metrics routes.

PR #42, opened 2021-04-17
Feature S3C-4300 stretched backbeat

vrancurel:
Wip


docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,344 @@
...
+
+Thanks to the settings above we can tolerate 2 servers failure without
+failing the writes but we sacrifice consistency (we can have
+inconsistencies in topic replicas limited to the buffer window).
+
+We chose that strategy as a trade-off for availability and consistency
+assuming inconsistency is very rare due to local proximity of servers.
+
+## What we shall do
+
```
vrancurel, line 130: @terrydervaux are we in <5ms or >= 5ms most of the time ?

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,344 @@
...
+Topic:backbeat-bucket-notification      PartitionCount:5        ReplicationFactor:3     Configs:max.message.bytes=5000000
+        Topic: backbeat-bucket-notification     Partition: 0    Leader: 2       Replicas: 2,3,4 Isr: 3,4,2
+        Topic: backbeat-bucket-notification     Partition: 1    Leader: 3       Replicas: 3,4,0 Isr: 3,4,0
+        Topic: backbeat-bucket-notification     Partition: 2    Leader: 4       Replicas: 4,0,1 Isr: 4,0,1
+        Topic: backbeat-bucket-notification     Partition: 3    Leader: 0       Replicas: 0,1,2 Isr: 0,1,2
+        Topic: backbeat-bucket-notification     Partition: 4    Leader: 1       Replicas: 1,2,3 Isr: 1,3,2
+```
+
+As we see we have 5 partitions by default, each having 3 replicas with
+an ISR of 2.
```
naren-scality, line 108: Isn't `ISR` 3 in the case above?

vrancurel: No ISR is 2 (see ./run-backbeat-queue/templates/server.properties.j2: `min.insync.replicas=2` )

cmrh (commented):
I think mirror maker gives us the best range of flexibility with regards to latency.

docs/design/backbeat-stretched.md
cmrh: I believe we should have a max latency we can handle value based on our data. More often than not, we do not know the latency for sure till we deploy. A max allowable latency if possible would be ideal

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
+# Backbeat Stretched
+
+## Summary
+
+S3C currently refuses to install backbeat on stretched deployments,
+unless the inter-site latency is < 5ms in which case the deployment
+can be forced at the risk and peril with
```
jonathan-gramain, line 7: at your own risk

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+While Metadata and Redis-sentinel being previously addressed, our
+focus is on Zookeeper and Kafka.
+
+Zookeeper is known to behave correctly with high latency but will
+require a new component called `Zookeeper Observer` for failing over
+correctly.
+
+In order to achieve geo-replication, Kafka could be configured with
+Replicas, ISR (In-Sync-Replicas) and rack-awareness but if not done
+correctly it could dramatically slows down the I/O performance.
```
jonathan-gramain, line 39: slow down

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+(including the partition-leader) is referred to as the ISR (In-Sync
+Replicas).
+
+The minimum number of in-sync replicas (`min.insync.replicas` in Kafka
+server.properties) specifies how many replicas that are needed to be
+available for the producer to successfully send records to a
+partition. 
+
+During a broker outage, all partition replicas on the broker become
+unavailable, so the affected partitions’ availability is determined by
```
jonathan-gramain, line 75: Does this only apply to partitions where the leader was on the broker that is down? Or is it also affecting follower replicas? Later you're saying "one of these in-sync replicas will become the interim partition leader" so I understand the explanation is only for when the leader fails, maybe could be made explicit.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+following values:
+
+* 0: return success at the moment the request is sent
+* 1: return success once the partition leader acknowledges the request
+* all: return success once all replicas in the ISR acknowledge the request
+
+There is another Kafka feature called Rack Awareness: Each broker is
+configured with a label describing which "rack" (or datacenter) the
+broker is within. Then, when Kafka assigns replicas across different
+brokers, it spreads the replicas across the available racks
```
jonathan-gramain, line 99: missing final `.`.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+# /opt/scality/s3tools/bin/backbeat-topic-info  | grep notif
+Topic:backbeat-bucket-notification      PartitionCount:5        ReplicationFactor:3     Configs:max.message.bytes=5000000
+        Topic: backbeat-bucket-notification     Partition: 0    Leader: 2       Replicas: 2,3,4 Isr: 3,4,2
+        Topic: backbeat-bucket-notification     Partition: 1    Leader: 3       Replicas: 3,4,0 Isr: 3,4,0
+        Topic: backbeat-bucket-notification     Partition: 2    Leader: 4       Replicas: 4,0,1 Isr: 4,0,1
+        Topic: backbeat-bucket-notification     Partition: 3    Leader: 0       Replicas: 0,1,2 Isr: 0,1,2
+        Topic: backbeat-bucket-notification     Partition: 4    Leader: 1       Replicas: 1,2,3 Isr: 1,3,2
+```
+
+As we see we have 5 partitions by default, each having 3 replicas with
```
jonathan-gramain, line 115: As Naren commented, the min-ISR comes from the configuration, it's not shown in the topic info, hence we could say instead "each having 3 replicas, and the Kafka brokers are configured with a min-ISR of 2" (for example)

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+an min-ISR of 2.
+
+In S3C we use the default value of `ack=1` which means that the replicas
+are not totally synchronously maintained. They are however potentially
+very close (messages within the buffer window are not required to be
+in-sync).
+
+Note: Setting `0` or `1` can lead to inconsistent partitions
```
jonathan-gramain, line 123: You may link to the article showing this possible behavior, which gives more details about this issue.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+(e.g. message differing for identical offsets if a leader vanishes
+temporarily and another leader is elected within the buffer
+window).
+
+Also since we use `keyed partitioning` writers error out when a partition
+is not available.
+
+Thanks to the settings above we can tolerate 2 servers failure without
```
jonathan-gramain, line 131: two server failures

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+servers.
+
+Note: The inconsistency shall be checked by the application itself
+(e.g. with an anti-entropy algorithm for CRR).
+
+### Availability Zone Latency scenario (<5ms)
+
+If we want to deploy Kafka stretched correctly and in AZ (availability
+zones) scenarios where latency is OK (<5ms), we need to have 6
+replicas per partition and an ISR of 4 with `ack=1` with
```
vrancurel, line 146: @jonathan-gramain actually if we have replicas=3 ISR=2 with rack-aware we could force at least 1 replica on the other site right ?

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+rack-awareness enabled. The first challenge is be to properly label
```
jonathan-gramain, line 147: "the first challenge is to"

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+the different kafka-brokers with a tool similar to Metadata
+`ToplogyGenerator`, then to properly assign the replicas of each
```
jonathan-gramain, line 149: I believe that with a recent version of kafka >= 2.4, the assignment derives directly from the labeling? i.e we do not have to do it ourselves, only the labeling (I may be wrong)

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+partition onto brokers belonging to different sites. The consumers
+also need to be configured (programmed! with a special selector) to
```
jonathan-gramain, line 151: Same, it might be automatic with a more recent version of kafka and consumers, to be checked.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+talk to the nearest brokers (same site) otherwise the overall reads
+can be slow.
+
+Note: By specifying `ack=1` the current state of the queue would be
+potentially sometimes inconsistent as we do today, but with the
```
jonathan-gramain, line 156: "as it is today"

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+additional negative constraint of the inter-site latency.
```
jonathan-gramain, line 157: "the additional negative effect"

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+
+Note: As a consequence, it is even more critical than the consistency
```
jonathan-gramain, line 159: Could be merged with the preceding "Note", and suggest rephrasing "it is even more critical that the consistency of the local replicas is maintained by the application"

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+
+![Dynamic Overview](img/backbeat-stretched/dynamic-overview-mirror-maker.png)
+
+| Step  | Description |
+| ------------- | ------------- |
+| 1. Get S3 Raft Log | Bucket Notification get the S3 Raft Log messages|
+| 2. Write event in queue | Bucket Notification writes the events in the topic|
+| 3. Write S3C Raft Log offset | The Queue Populator records the current S3 Raft log offsets.
+| 3bis. Write S3C Raft Log Offset as a "fake" event | Periodically record the state of S3C raft log offsets
+in the topic as a "fake" event for safe failover restart|
```
jonathan-gramain, line 375: In the formatted page, this line wraps around in the first column on a next row

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,432 @@
...
+-->
+
+![Overview](img/backbeat-stretched/overview-rack-aware.png)
+
+| Component  | Description |
+| ------------- | ------------- |
+| backbeat_* | Backbeat instances dispatched on all sites (e.g. `bucketNotifQueuePopulator` and `bucketNotifProcessor`) |
+| zookeeper_* | Zookeeper cluster nodes dispatched on all sites (1 per server) |
+| kafkaBroker_*_Rack* | Kafka cluster dispatched on all sites with Rack tag (1 per server) |
+| zookeeperObserver | Passive node listening to the messages and storing data in the ledger but not participating to the quorum |
```
jonathan-gramain, line 253: Wondering, wouldn't it be better to have 3 normal zookeeper nodes on one site, and 3 observers on the other site? So that the latency does not impact the elections process. The failover playbook would then have to set all 3 observers as participating nodes instead of just one.

docs/design/backbeat-stretched.md
jonathan-gramain: Why, because Kafka would enforce the ISR to always be across sites (racks)?

docs/design/backbeat-stretched.md
vrancurel: Yes I am talking about the proper labelling we have to do in Federation

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,431 @@
...
+the existence and status of their other replicas. If a partition has
+no additional replicas, the partition becomes unavailable.
+
+If a partition has additional replicas that are in-sync, one of these
+in-sync replicas will become the interim partition leader. Finally, if
+the partition has addition replicas but none are in-sync, we have a
+choice to make: either we choose to wait for the partition leader to
+come back online–sacrificing availability — or allow an out-of-sync
+replica to become the interim partition leader–sacrificing consistency
+(Source: see links below).
```
vrancurel, line 84: @jonathan-gramain do you know how can we make the choice practically ?

docs/design/backbeat-stretched.md
vrancurel: Ideally if we could force one copy on each site, like in Metadata that would be nice. Otherwise there is a high probability that the quorum will only be on the majority site.

One alternative is to have all 6 ZK nodes active and have a quorum of 4.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+don't want to lose the queue, to avoid lost/double notifications
+
+For 1. we don't really care about the rack-aware, number of copies in
+sync, since it is going to slow the perfomance down.
+
+For 2. we do, but we have only 1 broker per server, so we have to
+somehow find the best tradeoff between 1. and 2.  So if we configure
+rack-aware, type "1" workflows will benefit from it indirectly.
+
+The problem is that `min.insync.replicas` cannot be set programmatically
```
jonathan-gramain, line 207: Apparently it's a topic-level parameter, so it should be possible to configure a different one per topic.

See e.g. https://stackoverflow.com/questions/62326946/kafka-min-insync-replicas-interpretation

So we could consider using different settings for CRR and bucket notification topics, for example.

cmrh (approved):

rdebay-scality (approved):

docs/design/backbeat-stretched.md
vrancurel: That's right but let's not touch Backbeat code in this EPIC as much as possible.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+#### Failover (Temporary Minority Site Loss)
+
+* Remove leader enforcement in Metadata (see Metadata docs).
+* Change the bucket-notification topics replication factor to 3
+* Change the number of ZK nodes to 3
+
+#### Failover (Temporary Majority Site Loss)
+
+* Metadata specific playbook to promote warm-stand-by
+* Need special playbook for promoting the `zookeeperObserver` into a normal node
```
scality-gdoumergue, line 414: - We had a playbook to promote WSBs, but we depreciated it because maintaining took too much time. Instead, we use the DR tag, which uses the same roles and code path than "normal" topology configuration.
- It would be also useful to use the same code (i.e TopologyGenerator) to decide whether a zk observer needs to be promoted

docs/design/backbeat-stretched.md
jonathan-gramain: OK fair enough, but please correct the phrasing then:
```suggestion
We prefer not to change `min.insync.replicas` programmatically
```

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+* Change the bucket-notification topic replication factor back to 4
+* Change the number of ZK nodes to 5
+* Re-enable the ZK observer on WSB
+
+## Suggested Changes
+
+In all cases we shall review per workflow queue inconsistency issues: 
+
+* Anti-entropy for CRR since we use `ack=1`.
+* No problem with Bucket Notification since we plan to use `ack=all`.
```
jonathan-gramain, line 440: Please correct me if I'm wrong, just trying to validate my assumptions, here we will need to change backbeat code to configure the bucket notification producer with `ack=all`, correct? I.e. not a kafka parameter.

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+* Upgrading Kafka (e.g. to 2.4.0 for rack-aware consumption).
+* Deploy a Zookeeper Observer.
+* Tag the kafka-brokers with the rack (site ID).
+* Configure the proper replication factor for each topic (e.g. 3 for replication and 4 for bucket notification).
+* Modify Backbeat Bucket Notification to use `ack=all`.
```
jonathan-gramain, line 445: This line seems to confirm my assumption actually :slightly_smiling_face:

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+    Topic: backbeat-bucket-notification    Partition: 1    Leader: 0    Replicas: 0,4,2    Isr: 0,4,2
+    Topic: backbeat-bucket-notification    Partition: 2    Leader: 3    Replicas: 3,0,1    Isr: 3,0,1
+    Topic: backbeat-bucket-notification    Partition: 3    Leader: 1    Replicas: 1,3,4    Isr: 1,3,4
+    Topic: backbeat-bucket-notification    Partition: 4    Leader: 4    Replicas: 4,1,2    Isr: 4,1,2
+```
+
+For bucket notification we may set a replication factor of 4 which
+will dispatch 2 copies on 2 sites.
+
+Since we cannot change the `min.insync.replicas=2` there is no
```
jonathan-gramain, line 249: Why cannot we change this parameter, is it because of the increased latency? Could be good to add the reason

jonathan-gramain: Or is it because there is no one-size-fits-all for all use cases and we don't want to specify it per topic

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+
+#### Enabling a Zookeeper Observer on the WSB
+
+Although ZooKeeper performs very well by having clients connect
+directly to voting members of the ensemble, this architecture makes it
+hard to scale out to huge numbers of clients. The problem is that as
+we add more voting members, the write performance drops. This is due
+to the fact that a write operation requires the agreement of (in
+general) at least half the nodes in an ensemble and therefore the cost
+of a vote can increase significantly as more voters are added.
```
jonathan-gramain, line 291:
```suggestion
of a write can increase significantly as more voters are added.
```

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+    agent backbeat_5
+    agent zookeeper_5
+    agent kafkaBroker_5_Rack2
+    agent topic1Partition1Replica_5
+    agent topic1Partition2Replica_5
+  }
+  component wsb1Cluster1 {
+    agent backbeat_6
+    agent zookeeperObserver
+    agent kafkaBroker_6_Rack3
```
jonathan-gramain, line 347:
```suggestion
    agent kafkaBroker_6_Rack2
```

docs/design/backbeat-stretched.md
```diff
@@ -0,0 +1,610 @@
...
+  component md2Cluster1 {
+    agent backbeat_2
+    agent zookeeper_2
+    agent kafkaBroker_2_Rack1
+    agent topic1Partition1Replica_2
+    agent topic1Partition2Replica_2
+  }
+  component md3Cluster1 {
+    agent backbeat_3
+    agent zookeeperLeader
```
jonathan-gramain, line 370:
```suggestion
    agent zookeeper_3_Leader
```

docs/design/backbeat-stretched.md
jonathan-gramain: Ah I found where this comes from (https://zookeeper.apache.org/doc/r3.4.13/zookeeperObservers.html)

I wonder if it is a mistake on their side or I miss the point, in any case, a minor thing.

docs/design/backbeat-stretched.md
vrancurel: Not for now.

docs/design/backbeat-stretched.md
vrancurel: Yes I will modify the phrasing.

docs/design/backbeat-stretched.md
vrancurel: For now, one size fits all, and then we will do a review per workflow if necessary.

docs/design/backbeat-stretched.md
vrancurel: What we want is 5 ZK + 1 ZK observer.

jonathan-gramain (approved):

PR #43, opened 2021-04-19
feature: S3C-3669 bucket encryption design

jonathan-gramain:
Add a design document for the bucket encryption API support (bucket
and object-level).

### DIFF
--- a/docs/design/bucket-encryption.md
+++ b/docs/design/bucket-encryption.md
@@ -1,0 +1,384 @@
+# Bucket Encryption
+
+## Old support for bucket-level encryption
+
+### create_encrypted_bucket.js
+
+The previous way of encrypting objects in buckets is by using a script
+to create the bucket like:
+
+```
+docker exec scality-s3 node bin/create_encrypted_bucket.js -h 10.10.61.61 -p 8000 -a XIJG9OTSMLDDIEWX21HT -k BTNMVTDBujqpTUEVYVYtfhxExCWy3KvDlad4hmwC -b my-encrypted-bucket
+```
+
+The script is creating a bucket through the REST API, passing an extra
+proprietary header "x-amz-scal-server-side-encryption: AES256" to
+enable encryption using a newly created bucket master key stored in
+the KMS server, and referenced from the bucket metadata.
+
+It's also possible with the existing REST API to specify an existing
+master key ID to use by providing the extra
+"x-amz-scal-server-side-encryption-aws-kms-key-id" header, although
+the create_encrypted_bucket.js script does not provide this option.
+
+We plan to keep the compatibility with those headers for now but the
+compatibility will be removed eventually in favor of the bucket
+encryption API.
+
+## Bucket Encryption API
+
+We now support bucket encryption using the standard S3 APIs, to enable
+or disable encryption on existing buckets:
+
+- [PutBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketEncryption.html)
+
+- [GetBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketEncryption.html)
+
+- [DeleteBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucketEncryption.html)
+
+### PutBucketEncryption
+
+We support the following rule to enable encryption automatically on
+all objects in the bucket:
+
+```
+<ServerSideEncryptionConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+   <Rule>
+      <ApplyServerSideEncryptionByDefault>
+         <SSEAlgorithm>AES256</SSEAlgorithm>
+      </ApplyServerSideEncryptionByDefault>
+   </Rule>
+   ...
+</ServerSideEncryptionConfiguration>
+```
+
+Alternatively, we also allow specifying a master key ID to use for the
+bucket that will be sent when contacting the KMS via the
+KMSMasterKeyID tag, instead of letting S3C generate one automatically:
+
+```
+<ServerSideEncryptionConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+   <Rule>
+      <ApplyServerSideEncryptionByDefault>
+         <SSEAlgorithm>aws:kms</SSEAlgorithm>
+         <KMSMasterKeyID>abcdef</KMSMasterKeyID>
+      </ApplyServerSideEncryptionByDefault>
+   </Rule>
+   ...
+</ServerSideEncryptionConfiguration>
+```
+
+**Note**: only one `<Rule>` entry is allowed in the server-side
+encryption configuration.
+
+**Note**: the `<BucketKeyEnabled>` rule option is not supported.
+
+#### Implementation of PutBucketEncryption Regarding Bucket Metadata
+
+* If the `serverSideEncryption` does not exist, we create it and fill
+  with the relevant attributes (see [Internal Bucket
+  Metadata](#internal-bucket-metadata))
+
+* If the `serverSideEncryption` exists, we set its `mandatory`
+  attribute to `true`, and possibly set the `algorithm` to the new
+  configured algorithm, and if it is `aws:kms`, the **KMSMasterKeyId**
+  also gets stored in the `configuredMasterKeyId` attribute. We do not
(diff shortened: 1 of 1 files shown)


docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,333 @@
...
+  * store the encrypted data key and user-specified master key ID in
+    the object metadata. Do not create or alter the bucket server-side
+    encryption parameters.
+
+* if the algorithm value is **aws:kms** but no master key ID is
+  provided, follow the **AES256** logic above.
+
+## Bucket Policies
+
+**TODO**: determine if they are needed. They can prevent from changing
```
terrydervaux, line 156: As discussed on slack, we can remove bucket policies from our plan because our customers only want to prevent unencrypted objects in buckets.

tmacro (commented):
Might be worth it to include some examples of the `serverSideEncryption` section under a dedicated heading to directly show how the different "modes" effect the MD.

tmacro (approved):

jonathan-gramain:
@tmacro I reworked the internal bucket metadata section to contain separate examples for each case, it should be clearer (plus extra info on the API implementation too)

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,377 @@
...
+`serverSideEncryption` section, because it can be set or removed
+dynamically by the bucket encryption configuration API and should not
+alter the existing (permanent) bucket encryption managed key that is
+used to encrypt objects without an explicit KMS key ID.
+
+### Legacy Bucket Encryption Configuration (create_encrypted_bucket.js)
+
+For buckets that were created encrypted in previous releases using the
+`create_encrypted_bucket.js` script, the metadata format is the same
+than in section [Default Encryption Configuration With "AES256"
```
tmacro, line 299:
```suggestion
as in the section [Default Encryption Configuration With "AES256"
```

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,377 @@
...
+Algorithm](#default-encryption-configuration-with-aes256-algorithm).
+
+This way, we can treat is in the same way, simplifying the logic and
```
tmacro, line 302:
```suggestion
This way, we can treat it in the same way, simplifying the logic and
```

tmacro (approved):
Looks good, the various modes are much more clear now.

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,377 @@
...
+all objects in the bucket:
+
+```
+<ServerSideEncryptionConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+   <Rule>
+      <ApplyServerSideEncryptionByDefault>
+         <SSEAlgorithm>AES256</SSEAlgorithm>
+      </ApplyServerSideEncryptionByDefault>
+   </Rule>
+   ...
```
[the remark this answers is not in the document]
jonathan-gramain: Yes sounds good to me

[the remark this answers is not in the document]
jonathan-gramain: Maybe the code already checks this, but it should then return an error if there are multiple rules.

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,380 @@
...
+
+Alternatively, we also allow specifying a master key ID to use for the
+bucket that will be sent when contacting the KMS via the
+KMSMasterKeyID tag, instead of letting S3C generate one automatically:
+
+```
+<ServerSideEncryptionConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+   <Rule>
+      <ApplyServerSideEncryptionByDefault>
+         <SSEAlgorithm>aws:kms</SSEAlgorithm>
```
tmacro, line 63: I just discovered AWS allows a PutBucketEncryption with algo `aws:kms` and no specified `KMSMasterKeyID`. a subsequent GetBucketEncryption does not return a key id either.

jonathan-gramain: Yes, so we need to condition `aws:kms` **and** key provided in order to use a configured key and return it via GetBucketEncryption, correct? otherwise it's equivalent to AES256.

tmacro: Yes, I imagine it effects billing on AWS, but for our use its the same.

terrydervaux (approved):
It's super clear. Thanks @jonathan-gramain .

We will need to add in this design document how we manage the S3C configuration(federation) when S3C needs to be integrated with a multi-nodes KMS. It will be in another PR(S3C-4362).

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,380 @@
...
+We now support bucket encryption using the standard S3 APIs, to enable
+or disable encryption on existing buckets:
+
+- [PutBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketEncryption.html)
+
+- [GetBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketEncryption.html)
+
+- [DeleteBucketEncryption](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucketEncryption.html)
+
+### PutBucketEncryption
```
scality-gdoumergue, line 39: Do we already implement the BucketKeyEnabled rule? We should explicitely tell it in this doc.

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,380 @@
...
+  the bucket, when the user does not provide its own KMS key ID to use
+  in the "aws:kms" encryption mode (and when the default bucket
+  encryption does not contain an explicit one). It may also be the
+  hex-encoded master encryption key itself if there is no KMS server,
+  such as in this example (for testing purpose).
+
+**Note**: On AWS, the master key ID is an ARN of the form
+  "arn:aws:kms:region:account:key/uuid". S3C does not follow this
+  convention, in this example it is a "file" backend key (only used
+  for testing), and key IDs look different with a real KMS server.
```
scality-gdoumergue, line 243: Are we sure it won't break any compatibility with S3 clients?

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,380 @@
...
+
+* **cryptoScheme** is used to version the encryption method and
+  parameters used internally, like the "salt" value.
+
+* **cipheredDataKey** is the object data encryption key, itself stored
+  encrypted with the master key
+
+## Bucket Policies
+
+We currently do not plan to support bucket policies related to encryption.
```
scality-gdoumergue, line 363: Can you specify the name of the related conditions? 
I guess it's:
- s3:x-amz-server-side-encryption
- s3:x-amz-server-side-encryption-aws-kms-key-id

scality-gdoumergue (changes requested):
Some details to be added.
I would also like to make KMS configuration easier. Today we have to download a .properties file and upload it into Federation. I think it'd be cool to have it directly integrated as a template, with a new Ansible variable set. Do you think it's feasible to add it to this design doc?

docs/design/bucket-encryption.md
jonathan-gramain: I don't know about this, but this part will be unchanged compared to what already exists, so assuming clients work with the current scheme, they should still work with the new API support.

If it indeed does break the compatibility for certain clients that we would not have seen yet, it's a new issue and the best would be to tackle it as a separate bug ticket.

docs/design/bucket-encryption.md
jonathan-gramain: Yes, or only `s3:x-amz-server-side-encryption` should be enough for the purpose, I will add a precision.

docs/design/bucket-encryption.md
```diff
@@ -0,0 +1,384 @@
...
+        "cipheredDataKey": "d88SLogLTw8nzRyO0Tf0qynQPLubT0N9mBDAxEZ5ww0="
+    }
+]
+```
+
+* **cryptoScheme** is used to version the encryption method and
+  parameters used internally, like the "salt" value.
+
+* **cipheredDataKey** is the object data encryption key, itself stored
+  encrypted with the master key
```
scality-gdoumergue, line 361: Does it mean that, during the object's download, S3C doesn't need to contact the KMS server to decrypt the object?

scality-gdoumergue (approved):

PR #44, opened 2021-05-05
CRR metrics design

miniscruff:
Will need to use `mkdocs serve` to view the mermaid sequence diagram. 

image

Status processor not shown.

Note I only included information on metrics specific to CRR and not generic metrics we might want to include everywhere. An example could be HTTP errors, NodeJS statistics or whatever.


docs/design/crr-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+replication_processor_bytes | Number of bytes replicated not including metadata | Counter
+replication_processor_elapsed_seconds | Replication jobs elapsed time in seconds | Histogram
+
+The difference between processor objects and queued objects is the current number of waiting objects.
+The difference between processor bytes and queued bytes is the current amount of bytes waiting to replicate.
+
+## Labels
+| Name | Description |
+| ---- | ----------- |
+origin | "lifecycle"
```
miniscruff, line 50: @jonathan-gramain I looked around and found this value is hardcoded as lifecycle, I would like to know if there is any context around this I can include or of course if I am wrong and this isn't the case.

docs/design/crr-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+## Metrics
+| Name | Help | Type |
+| ---- | ---- | ---- |
+replication_populator_messages | Total number of kakfa messages produced by the queue populator | Counter
+replication_populator_objects | Total objects queued for replication | Counter
+replication_populator_bytes | Total number of bytes queued for replication not including metadata | Counter
+replication_processor_messages | Total number of kakfa messages consumed by the queue processor | Counter
+replication_processor_objects | Number of objects replicated | Counter
+replication_processor_bytes | Number of bytes replicated not including metadata | Counter
+replication_processor_elapsed_seconds | Replication jobs elapsed time in seconds | Histogram
```
NicolasT, line 42: I don't think the *name* of a metric should represent the 'application' that's being metered, except in very specific cases. This belongs as a label.

Think of it like this: in essence, both the populator and the processor work on queues of items. So, the metrics can be the same. Essentially, you should be able to create a single Prometheus query (or Grafana dashboard) in which you, for example, measure the rate at which objects are processed (added to the queue by the populator, or dequeued by the processor, that makes no difference). Then, optionally, you can add a label selector to the query to focus on one or the other.

I hope that's somewhat clear.

It's similar to HTTP servers, for example. There are a bunch of 'de-facto standard' Prometheus metrics for anything that somewhat resembles an HTTP server. You can design a single dashboard which works for both Nginx as well as httpd. And if you have both running in your environment, the metrics with the same 'name' can be discerned by adding a label selector to the queries, focusing on a single source of metrics (nginx or httpd, for example).

miniscruff: I see, https://prometheus.io/docs/practices/naming/ does have a caveat for when metrics are more generic. The elapsed seconds one is the only one that is "processor" specific as it tracks the total time from in Kafka to complete. Removing "replication" from the metric names and replacing it with an "appName" ( or equivalent if there is some standard that would work of course ) would be fine. 

I don't see a problem with this, curious of what @terrydervaux has to say.

NicolasT: Also CC @rahulreddy @RachedBenMustapha @vrancurel We discussed this before in chapter and 1x1, remember?

miniscruff:
Need to add status processor

docs/design/crr-metrics.md
jonathan-gramain: For lifecycle transitions, the copy (replication) mechanism might be used for other features that need to copy data like CRR, in the future. IIRC, this field was to distinguish which high-level feature is using the replication, hence only "lifecycle" for now, but potentially others later.

miniscruff: ok that is what Rahul sort of mentioned, if you could propose a description I can include it, otherwise I will see what I come up with later on.

docs/design/crr-metrics.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Queue Processor Flow
+```mermaid
+sequenceDiagram
+    participant Kafka
+    participant proc as Queue Processor
+    participant to as To Location
+    participant Prometheus
+    proc-->>Kafka: Pull message
+    proc->>to: Send data to new location
```
jonathan-gramain, line 26: You will also likely need to track the read side of the copy, i.e metadata + data operation on source cloudserver.

docs/design/crr-metrics.md
jonathan-gramain: Could be "Origin feature related to the metric"

docs/design/crr-metrics.md
miniscruff: Gotcha, will add this

docs/design/crr-metrics.md
```diff
@@ -0,0 +1,126 @@
...
+    end
+    Prometheus-->>proc: Scrape metrics
+```
+
+### Metrics
+
+| Name | Help | Type |
+| ---- | ---- | ---- |
+replication_objects_status_changed | Number of objects updated | Counter
+kafka_lag | Number of kafka objects we are behind in the queue | Gauge
```
jonathan-gramain, line 109: Having `kafka_messages_consumed` here could make sense too, unless it duplicates `replication_objects_status_changed`

docs/design/crr-metrics.md
```diff
@@ -0,0 +1,126 @@
...
+    Prometheus-->>proc: Scrape push metrics
+```
+
+### Metrics
+
+| Name | Help | Type |
+| ---- | ---- | ---- |
+replication_processed_objects | Number of objects replicated | Counter
+replication_processed_bytes | Number of bytes replicated not including metadata | Counter
+replication_processor_elapsed_seconds | Replication jobs elapsed time in seconds | Histogram
```
jonathan-gramain, line 71: As a suggestion, we could consider labeling by ranges of size, e.g. to tell how long small objects take to be replicated vs. big ones. But it's already useful as is, just thinking for a potential future improvement.

jonathan-gramain (approved):

docs/design/crr-metrics.md
miniscruff: It would be a duplicate as replication objects status changed is the end result of reading a kafka message.

naren-scality (approved):

PR #46, opened 2021-05-17
doc: init S3 lifecycle expiration requirements

terrydervaux:
Lifecycle expiration requirements for S3 connector.

### DIFF
--- a/docs/requirements/s3-lifecycle-expiration.md
+++ b/docs/requirements/s3-lifecycle-expiration.md
@@ -1,0 +1,104 @@
+# S3 Lifecycle Expiration
+
+This document describes S3 Lifecycle expiration requirements for S3 connector.
+
+## Context
+
+Scality customers wants to automate the deletion of objects and object versions. 
+They want to have the ability to configure the expiration using a Graphical User 
+Interface(GUI) or the standardized AWS S3 API `s3:PutBucketLifecycleConfiguration`,
+`s3:DeleteBucketLifecycle`, and `s3:GetBucketLifecycleConfiguration`.
+
+## Roles
+
+Storage Administrator and the Storage Account Owner roles are impacted by the lifecycle
+expiration feature.
+
+### Storage Account Owner Role
+
+Storage account owners are responsible for configuring lifecycle expiration for an
+S3 client application.
+
+### Storage Administrator Role
+
+Storage Administrators are responsible for operating the Scality platform. They
+monitor the status and performance of the components that deliver the lifecycle
+expiration service.
+
+## User Stories
+
+### Storage Administrator Story
+
+**As a** Storage Administrator
+
+**I want to** a monitor the lifecycle expiration components
+
+**So that** I can identify configuration, performance, or network issues
+
+### Storage Account Owner Story
+
+**As a** Storage Account Owner
+
+**I want to**  delete objects or previous version of an object after a given amount of
+days using a lifecycle expiration configuration
+
+**In order to** store cost-effectively my objects
+
+## Acceptance Criteria
+
+### Deployment and Configuration
+
+The expiration system must:
+
+- work on 1-site, 2-site and 3-site stretch deployments
+- have the ability to scale horizontally
+
+### Performances
+
+The expiration system must be able to handle:
+
+- 20 000 buckets.
+- 80 million buckets with few objects.
+- 10 billion objects within a bucket.
+
+### API
+
+#### Lifecycle Expiration Configuration
+
+Lifecycle expiration configuration must:
+
+- be configurable from the endpoint that is used to create, read, update, and
+  delete (CRUD) objects and buckets.
+- expire object and object versions.
+- allow filtering objects or object versions by key name prefix, object tags, or
+  a combination of both.
+- be able to delete incomplete multipart uploads
+- respect object lock configuration
+- work on objects replicated from another RING
+- work on objects replicated with an object lock configuration from another RING
+
+#### Monitoring API
+
+Lifecycle expiration components must be exposed monitoring metrics through HTTP
+endpoint following Prometheus standards.
+
+### Identity and Access Management
+
+IAM policies must support deny or allow `s3:PutBucketLifecycleConfiguration`,
+`s3:DeleteBucketLifecycle`, and `s3:GetBucketLifecycleConfiguration` actions.
+
+### UI
+
+#### Lifecycle Expiration Configuration UI
+
+Lifecycle expiration configuration must be configurable through the S3 browser.
+
+#### Monitoring UI
+
+Lifecycle expiration components must have a Grafana dashboard to troubleshoot
+performance, configuration, and/or network issues.
+
+### Sizing
+
+The RING sizing tool must be updated to take into consideration the lifecycle
+expirations components requirements(RAM, CPU,and storage consumption).
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -18,6 +18,7 @@ extra_javascript:
 nav:
   - Requirements:
     - 'S3 Replication Monitoring': 'requirements/s3-replication-monitoring.md'
+    - 'S3 Lifecycle Expiration': 'requirements/s3-lifecycle-expiration.md'
   - Design:
     - 'General': 'design/index.md'
     - 'Bucket Notifications': 'design/bucket-notification.md'


docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+- allow filtering objects or object versions by key name prefix, object tags, or
+  a combination of both.
+- be able to delete incomplete multipart uploads
+- respect object lock configuration
+- work on objects replicated from another RING
+- work on objects replicated with an object lock configuration from another RING
+
+#### Monitoring API
+
+Lifecycle expiration components must be exposed monitoring metrics through HTTP
```
RachedBenMustapha, line 82: This does not currently exist for expiration and needs to be developed. Since it was not part of the original epic's acceptance criteria, it needs to be tracked there as well.

A list of which metrics have to be exposed should be part of this requirements document.

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+**In order to** store cost-effectively my objects
+
+## Acceptance Criteria
+
+### Deployment and Configuration
+
+The expiration system must:
+
+- work on 1-site, 2-site and 3-site stretch deployments
+- have the ability to scale horizontally
```
RachedBenMustapha, line 54: To the extent that the customer's metadata cluster is set up for horizontal scaling, as it will be the bottleneck once backbeat workers and kafka brokers/partitions have been scaled up.

Does this horizontal scaling requirement also apply to 2- and 3-site stretch?

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Performances
+
+The expiration system must be able to handle:
```
RachedBenMustapha, line 58: Handle in how much time? What's the acceptable resource overhead? What's the SLA tolerance?

docs/requirements/s3-lifecycle-expiration.md
xaf-scality: Yes also apply to 2- and 3site stretch. We already have 3-site customer with huge day to day load of Lifecycle expiration. So need to horizontal scale and support one site lost.

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+- be configurable from the endpoint that is used to create, read, update, and
+  delete (CRUD) objects and buckets.
+- expire object and object versions.
+- allow filtering objects or object versions by key name prefix, object tags, or
+  a combination of both.
+- be able to delete incomplete multipart uploads
+- respect object lock configuration
+- work on objects replicated from another RING
+- work on objects replicated with an object lock configuration from another RING
+
```
xaf-scality, line 79: - work on objects in an encrypted bucket

xaf-scality (changes requested):

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+### API
+
+#### Lifecycle Expiration Configuration
+
+Lifecycle expiration configuration must:
+
+- be configurable from the endpoint that is used to create, read, update, and
+  delete (CRUD) objects and buckets.
+- expire object and object versions.
```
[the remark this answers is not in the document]
scality-gdoumergue: In fact I tested AWS and the official doc contradicts the reality. When I have this kind of object:
- Two noncurrent versions
- One DeleteMarker as current version

AWS deletes the 3 versions with this lifecycle policy:

{
    "Rules": [
        {
            "ID": "Expire",
            "Status": "Enabled",
	    "Prefix": "",
            "NoncurrentVersionExpiration": 
                {
                    "NoncurrentDays": 1
                },
            
            "Expiration": 
                {
                    "ExpiredObjectDeleteMarker": true
                }
        }
    ]
}

So, disregard my latest comment.

docs/requirements/s3-lifecycle-expiration.md
rahulreddy: Supporting stretched failover and failback for Backbeat services will be followed up by #42 . Given that Backbeat processes are stateless they inherently support scaling horizontally. I understand that for optimal performance will depend on a few factors such as available CPU/RAM resources for the workers, Metadata performance etc.

docs/requirements/s3-lifecycle-expiration.md
rahulreddy: I am moving this out since it will be addressed in a future version.

docs/requirements/s3-lifecycle-expiration.md
rahulreddy: I am adding to the list. @RachedBenMustapha @alexanderchan-scality @nicolas2bert Is this already considered or will this be an out of scope change?

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+  a combination of both.
+- be able to delete incomplete multipart uploads
+- respect object lock configuration
+- work on objects replicated from another RING
+- work on objects replicated with an object lock configuration from another RING
+
+#### Monitoring API
+
+Lifecycle expiration components must be exposed monitoring metrics through HTTP
+endpoint following Prometheus standards.
```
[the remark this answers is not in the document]
rahulreddy: I am adding comment that metrics will be followed up in a future release.

docs/requirements/s3-lifecycle-expiration.md
```diff
@@ -0,0 +1,104 @@
...
+### UI
+
+#### Lifecycle Expiration Configuration UI
+
+Lifecycle expiration configuration must be configurable through the S3 browser.
+
+#### Monitoring UI
+
+Lifecycle expiration components must have a Grafana dashboard to troubleshoot
+performance, configuration, and/or network issues.
```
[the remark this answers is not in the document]
rahulreddy: I am adding comment that metrics will be followed up in a future release.

docs/requirements/s3-lifecycle-expiration.md
RachedBenMustapha: We have not explicitly been testing with encrypted buckets (given that the feature is being developed at the same time as lifecycle). I wouldn't anticipate that encrypted buckets would need special treatment but will schedule time to ensure interoperability.

rahulreddy: Yeah I am not anticipating surprises either since we are using S3 api for the actions. The encrypted buckets meant either using what we had before as bucket level encryption or the new apis, since the backend code did not change.

scality-gdoumergue (approved):

scality-gdoumergue:
Excerpt from [AWS doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html):

> Disabling or deleting Lifecycle rules
> 
> When you disable or delete Lifecycle rules, Amazon S3 stops scheduling new objects for deletion or transition after a small delay. Any objects that were already scheduled are unscheduled and are not deleted or transitioned.

Do we also do that? For example, if the LOP has a 1M events lag, and the customer disables a particular rule, will the LOP continue to delete the objects that are still in the topic? If not, we should explicitely say it in the requirements and the official doc.

docs/requirements/s3-lifecycle-expiration.md
rahulreddy: So we are following AWS behavior in the sense the objects are deleted in the span of less than 1-2 days. This of course depends on the system ensuring that there is enough system resources (CPU, RAM, I/O) to handle the lifecycle workload. How much we have to size for we will know from the outcome of perf testing.

rahulreddy:
> Do we also do that? For example, if the LOP has a 1M events lag, and the customer disables a particular rule, will the LOP continue to delete the objects that are still in the topic? If not, we should explicitely say it in the requirements and the official doc.

@scality-gdoumergue We are not doing that today. There will be a significant impact on the velocity of deletes as it incurs additional requests to Metadata to check the lifecycle status of the bucket. However we need to figure out a way to do this in a smarter way. I added that ticket for the follow up epic Lifecycle optimization. We are going to document this as a limitation in 8.5 release.

PR #49, opened 2021-05-27
test: S3C-4409 bucket encryption test plan

jonathan-gramain:
- a deployment test with Federation

- a functional test to check encryption correctness

- a set of performance tests with cosbench to execute on a perf lab

This was inspired from https://github.com/scality/artesca/blob/development/1.0/docs/developer/test-plan/data-services.rst, ty @RachedBenMustapha !


tmacro (approved):

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Deployment
+
+### Deployment through Federation
+
+#### Actions
+
+* Deploy the KMS appliance to test in a single instance using [these instructions]
+  (https://documentation.scality.com/S3C/7.9.0/installation/install_s3c/Configuring_the_S3_Cluster/additional_features/encryption+key_mgmt_configuration.html).
+
```
scality-gdoumergue, line 11: We should configure the gemalto plugin to log into its own log file (/logs/gemalto.log). This can help us know that everything works fine.

scality-gdoumergue (changes requested):

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,212 @@
...
+  ```
+
+  * Create an empty object `empty-obj` in the bucket
+  * Create a 1KB object `1kb-obj` in the bucket with random data
+  * Create a 1MB object `1mb-obj` in the bucket with random data
+  * Create a 1GB MPU object `1gb-obj` in the bucket with random data
+  * Do a HEAD on each object
+  * Do a GET on each object
+
+#### Expected Results
```
scality-gdoumergue, line 57: Add an upgrade section.
- Install S3C 7.9
- Create an encrypted bucket the old way
- Upgrade to 7.10
- Encryption must still work for the old bucket.

scality-gdoumergue (changes requested):

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,212 @@
...
+      --server-side-encryption-configuration file://encryption.json
+  ```
+
+  * Create an empty object `empty-obj` in the bucket
+  * Create a 1KB object `1kb-obj` in the bucket with random data
+  * Create a 1MB object `1mb-obj` in the bucket with random data
+  * Create a 1GB MPU object `1gb-obj` in the bucket with random data
+  * Do a HEAD on each object
+  * Do a GET on each object
+
```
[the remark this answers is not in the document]
scality-gdoumergue: + Check on Ciphertrust web page that the s3-encrypted-bucket-1 key has been created.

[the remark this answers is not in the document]
scality-gdoumergue: Delete the objects, remove the bucket. The s3-encrypted-bucket-1 key on Ciphertrust should be removed as well.

scality-gdoumergue (changes requested):

docs/test-plan/bucket-encryption.md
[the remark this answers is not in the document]
jonathan-gramain: > Check on Ciphertrust web page that the s3-encrypted-bucket-1 key has been created.

Good idea. In practice I see that the keys are named with random strings that are stored in bucket metadata, like `ks-abcdef....`. So it's possible to check that the matching key exist from the bucket metadata, with a lookup of bucket metadata first.

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,212 @@
...
+#### Expected Results
+
+* Bucket encryption configuration was successfully applied
+* Each object metadata from the HEAD request shows that it is encrypted using AES256.
+* Each object data from the GET matches what has been put.
+
```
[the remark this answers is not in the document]
jonathan-gramain: The set of operations happening inside the KMS look rather obscure for now, but ideally we should be able to "decipher" what's going on. I'll just mention that there should be no error.

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,212 @@
...
+    * PUT with encryption compared to PUT without encryption
+    * GET with encryption compared to GET without encryption
+
+* There should be no or very few errors in cloudserver logs, errors
+  should be explained if any (e.g. network is saturated)
+* Observe the KMS cluster logs, there should be no error, or errors
+  should be identified and explained
+
+## Failure Tests
+
```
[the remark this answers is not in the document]
jonathan-gramain: The bucket creation is expected to fail, added a test for that

docs/test-plan/bucket-encryption.md
jonathan-gramain: Is it possible now or does it require code changes?

scality-gdoumergue: No code change required. Add these config lines in ProtectAppICAPI.properties:

Log_File=/logs/gemalto.log
Log_Level=HIGH

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,261 @@
...
+* Fetch bucket metadata (e.g. from the raft log), look at the
+  `serverSideEncryption.masterKeyId` value, check on the web interface
+  of the CipherTrust app, under "Key & Access Management -> Keys" that
+  a key named `ks-{{masterKeyId}}` has been created
+* Sproxyd keys contain a different (encrypted) payload than the
+  payload of the uploaded object
+* Cloudserver and Gemalto plugin logs show no error nor warning
+* KMS logs (in `/opt/keysecure/logs/keysecure.system.log`) show no
+  error during operations
+
```
scality-gdoumergue, line 79: Add a section to test KMS on versioned buckets.

scality-gdoumergue, line 79: Add a section to test encryption-related IAM authorization:

- The root user can always use the bucket-encryption related actions,
- An IAM user can not use the bucket-encryption related actions,
- An IAM user with the proper policy should be able to use the bucket-encryption related actions,
- Least-privilege must be enforced. If a user i attached to a policy that explicitely forbids the use of bucket-encryption related actions, another policy shouldn't be able to allow them.

jonathan-gramain: Added a versioned bucket as a new test case in this test

jonathan-gramain: Added the first three points.

For the fourth one, what kind of other policy that could allow the encryption actions are you referring to, bucket policies? IAM group policies? I know IAM has the notion of "Deny" policies for that, although I'm not familiar enough with their use to be comfortable describing how to set them in the right context, so help would be appreciated.

jonathan-gramain:
@scality-gdoumergue could you please review the latest changes? If good for you I'd like to merge this during this sprint (so this week).

Thanks!

docs/test-plan/bucket-encryption.md
jonathan-gramain: I realized that the ProtectApplCAPI.properties file seems to be used only when we setup the scality KMS with the NAE protocol, where the "scality-kms" container is used.

I did not test the NAE mode, only KMIP, and I understood from the PO that we were going to deprecate NAE and going to focus on KMIP only, hence we can probably drop this.

I believe the logging is configured directly in the Thales CipherTrust GUI when using KMIP.

I am not sure I understand fully how it is supposed to be configured, so please correct me if I'm wrong.

jonathan-gramain: I will remove the section that I added about logging then

docs/test-plan/bucket-encryption.md
```diff
@@ -0,0 +1,612 @@
+# Bucket Encryption Test Plan
+
+## Deployment
+
+### Deployment through Federation
```
jonathan-gramain, line 5: @scality-gdoumergue the deployment section was previously referring to how to deploy the "kms" (NAE) protocol, which we want to deprecate as I explained for the logging. I think there exists some documentation for kmip deployment but I could not find it. For now I added the steps that I took to deploy and configure the CipherTrust VM on the lab, but ideally we should have an up-to-date link on how to configure with KMIP.

docs/test-plan/bucket-encryption.md
scality-gdoumergue: @jonathan-gramain I don't understand.
- Deprecating NAE: what about the customers who already use NAE?
- Do you mean going full KMIP we don't need the code from the scality-kms container?

docs/test-plan/bucket-encryption.md
scality-gdoumergue: Create an IAM policy with:

`"Action":["s3:PutEncryptionConfiguration"],"Effect":"Deny"`

And a second one with:

`"Action":["s3:PutEncryptionConfiguration"],"Effect":"Allow"`

Attach the two policies to a single IAM user.

The user shouldn't be able to `aws s3api put-bucket-encryption`.

docs/test-plan/bucket-encryption.md
jonathan-gramain: > Deprecating NAE: what about the customers who already use NAE?

I believe it should still work as it used to, just that we are not focusing on testing it, we are focusing on KMIP.

> Do you mean going full KMIP we don't need the code from the scality-kms container?

Yes, that's correct, KMIP does not require the scality-kms container.

scality-gdoumergue (approved):

PR #52, opened 2021-07-27
S3C-4703: add lifecycle expiration test plan

alexanderchan-scality:
Added test plans:

- [ ] Deployment
- [ ] Upgrade Tests
- [ ] Functional
- [ ] Performance:
  - IOPS
  - 20K buckets
  - 10B objects
- [ ] Failure Tests:
  - Single bucket processor failure
  - Multi bucket processor failure
  - Single object processor failure
  - Multi object processor failure
  - Recovery tests


docs/test-plan/lifecycle-expiration.md
```diff
@@ -0,0 +1,436 @@
...
+  ```
+
+#### Expected Results
+
+* After a short time period, `expire-mpu-bucket` must not contain any
+  incomplete multipart uploads when performing mpu listing operation
+  ```bash
+    aws s3api list-multipart-uploads --bucket expire-mpu-bucket
+  ```
+
```
jonathan-gramain, line 293: I would suggest adding some functional tests related to filtering by prefix/tag.

Also, ideally we should check that lifecycle does *not* touch objects that are younger than the specified age, but since the granularity is the day it would require setting up the test the previous day. I believe it's okay since those tests are manual and can take extra time to complete. It may be done at the same time than the rest of the existing tests, by adding another bucket with the same policies but expiring after two days (so it should actually wait for one day due to the test flag being on)

docs/test-plan/lifecycle-expiration.md
```diff
@@ -0,0 +1,436 @@
...
+  ```
+* (TODO?: Run a cosbench workload in the background to simulate a normal operating load)
+
+#### Expected Results
+
+* The time to expire all the target objects should be within a reasonable
+  amount of time:
+    * measure the time to expire all the objects across the 20K buckets.
+
+### With large number of objects (10B objects)
```
jonathan-gramain, line 408: I am not convinced putting 10B objects is practical to do on a lab in a reasonable time :slightly_smiling_face: (but 10M surely is! Maybe even 100M)

PR #54, opened 2021-07-29
ft(UTAPI-2): Update utapi design to improve performance and disk usage

tmacro:


mrepop (approved):

saitharunthodupunuru (approved):

anurag4DSB (approved):

docs/design/utapi.md
```diff
@@ -57,216 +74,411 @@ pair `labels`, and an associated value.
...
+signifying the end of the time range of included checkpoints. The latest master
+snapshot is searched for and all checkpoints with timestamp lying between the
+master snapshot's timestamp and the passed timestamp are fetched. If no previous
+master snapshots are found a timestamp of `0` is used causing all checkpoints
+to be included. The checkpoints are then iterated over and summed together based
+upon their labels using the values from the previous snapshot as a base.
 
-For example the two checkpoints from the previous example:
+For example the three checkpoints from the previous example will produce snapshots
+holding the same values.
```
vrancurel, line 284: It would clearer to provide an example where it is not the same value for snapshots ? e.g. with a previous state for the bck's and acc.

docs/design/utapi.md
```diff
@@ -57,216 +74,411 @@ pair `labels`, and an associated value.
...
+      opId: "putObject",
 }
 ```
 
 Each event can also have attached metadata such as the `account` , `bucket`,
-or `location` of the affected resource. These are attached as Warp 10 `labels`
-to the event and are used during checkpoint creation to create the various
-levels of metrics.
+or `location` of the affected resource. These are used during checkpoint
+creation to create the various levels of metrics.
```
vrancurel, line 164: 1. That would be really nice not to pollute the global redis-local namespace (because it is used in other components) and as we saw on the field, there are sometimes millions of keys in the global namespace.

2. Also I would suggest to directly write the events in the shards, because I don't see the value of writing them first and then referencing them in the shards.

Hopefully implementing 2. will automatically solve 1.

docs/design/utapi.md
```diff
@@ -57,216 +74,411 @@ pair `labels`, and an associated value.
...
+utapi -> warp10 : Request metrics at timestamp from Warp 10
+activate warp10
+warp10 <-- warp10 : Load newest snapshot older than given timestamp for each available nodeId
+warp10 <-- warp10 : Load all checkpoints that lie between a nodeId's snapshot and given timestamp
+warp10 <-- warp10 : Load any late checkpoints from utapi.repair.checkpoint in the same range
+warp10 <-- warp10 : Load latest reindex record labeled with nodeId
+utapi <- warp10 : Return computed metrics
+deactivate warp10
+@enduml
+-->
```
vrancurel, line 481: We need to add a section on how we can guarantee the results are accurate to a pre-defined granularity:

1. we shall somehow indicate that a result is *transient*, e.g. if we query Utapi "live" result, it is very likely that the datalogs haven't propagated yet and the result is wrong.
2. we shall have a mode to query from a safe point, e.g. the last checkpoint. If we can evaluate the propagation time of the datalog, we can be sure that the result in the last checkpoint is safe. Similarly if we detect that the lag is greater than the last checkpoint, then we must return that somehow the result in "unsafe".

Please specify how we can achieve the 2 above points.

vrancurel (changes requested):

cmrh (approved):

docs/design/utapi.md
```diff
@@ -57,216 +74,411 @@ pair `labels`, and an associated value.
...
 
 #### Events
 
-Utapi stores four integers, using Warp 10's multivariate data type, for every
-operation it ingests, `objectDelta`, `bytesDelta`, `ingress`, and `egress`.
-They reflect how an operation affects the computed metric counters and can be
-both positive and negative.
+Utapi stores a protobuf object using Warp 10's binary data type containing
```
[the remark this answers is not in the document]
tmacro: Only the checkpoints are stored. I meant for `operation` to refer to checkpoints, but it doesn't convey that very well.

docs/design/utapi.md
vrancurel: So each stateful node can emit regularly a "canary". We can use this "canary" to determine the "freshness" of the result.

vrancurel: So the canary must be emitted by the stateful nodes every 15mn by default (tunable). A manual request could then establish the staleness or the lag of the datalog replication.

For data loss detection. When a Cloudserver cannot write any more to redis-local it emits a 'potential_dataloss' event directly to Warp10, then cache in its RAM. When it cannot write any more in its RAM it writes a 'dataloss' event directly to Warp10. A manual request could then detect data loss if a 'dataloss' or a 'potential_dataloss' event is not followed by a 'reindex'.

tmacro: I suggest adding a 3rd event to be published if the cache has successfully been flushed and no events are lost. We can use this to avoid reducing confidence in the results if no data have actually been lost.

vrancurel (approved):

